PR #4, opened 2015-06-21
Fixed model-collection and tests

jylauril:
* Fixed restful-model-collection's delete method as it was calling request with wrong params
* Adjusted restful-mode-collection-spec accordingly
* Added missing devDependencies to package.json


bengotow:
Thanks Jyrki—this looks great. Thanks for updating the specs!

PR #5, opened 2015-06-23
Extended events model to include creating and updating of model

jylauril:
* Added participant model to be used in the event's participants collection
* Added save method and missing attributes for the event model
* Created event-spec to cover the event model tests


bengotow:
Looks good to me—thanks for including tests!

PR #7, opened 2015-06-26
Add delta streaming API support

ericyhwang:
@bengotow 

- New `namespace.delta` to expose delta API
  - `namespace.delta.generateCursor(timestampMs, [callback])` to get a Nylas delta cursor from a C-like timestamp
  - `namespace.delta.startStream(cursor, excludeTypes = [])` to initiate a delta streaming API connection. It returns a `DeltaStream` which emits `response`, `delta`, and `error` events.
- Jasmine tests for delta streaming, covering:
  - Starting and closing a stream
  - Parsing the streaming response into JSON objects
  - Handling JSON objects broken into multiple data packets
  - Automatically restarting the connection after a timeout not receiving any data (normally the Nylas server sends a heartbeat every second on the streaming connection)
- NylasConnection updates
  - NylasConnection now clones the passed-in options and adds the standard Nylas options on the clone, instead of modifying the original passed-in object in-place.
  - Moved said option-modification to new `NylasConnection.requestOptions(opts)` method to build an option object suitable for passing to `request`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,6 +196,40 @@ namespace.drafts.find(savedId).then(draft) {
 
 ```
 
+Using the Delta Streaming API
+------
+
+```javascript
+var DELTA_EXCLUDE_TYPES = ['contact', 'calendar', 'event', 'file', 'tag'];
+var timestampMs = Date.now();
+
+namespace.deltas.generateCursor(timestampMs, function(cursor) {
+
+  // Save inital cursor.
+  persistCursor(cursor);
+
+  // Start the stream and add event handlers.
+  var stream = namespace.deltas.startStream(cursor, DELTA_EXCLUDE_TYPES);
+
+  stream.on('delta', function(delta) {
+    // Handle the new delta here.
+    console.log('Received delta:', delta);
+    // Save new cursor so this delta doesn't need to be re-fetched for future streams.
+    persistCursor(delta.cursor);
+
+  }).on('error', function(err) {
+    // Handle errors here, such as by restarting the stream at the last cursor.
+    console.error('Delta streaming error:', err);
+
+  });
+
+  // Closing the stream explicitly, if needed
+  stopButton.addEventListener('click', function() {
+    stream.close();
+  });
+})
+```
+
 Contributing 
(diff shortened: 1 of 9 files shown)


ericyhwang:
Not done yet - updating README.md with examples of using the delta streaming API. I'll do it once we're both happy with the state of the API.

bengotow:
Hey Eric—this looks great. Thanks for writing the exponential backoff logic into the `DeltaStream`. The tests look great as well. If you haven't already, can you sign the contributor license agreement (https://nylas.com/cla)? Then we can get this merged!

I think it'd be great to get a few examples in the README but it sounds like you're on top of that.

One minor nit: could we change the `delta` object to be `deltas`?

```
namespace.deltas.generateCursor(timestamp)
namespace.deltas.startStream(cursor, excludeTypes = [])
```

I think that'd make it match the work going on in our other SDKs a bit better.

ericyhwang:
Rename done as requested, and added an example to README.md. PTAL.

ericyhwang:
Quick note - I currently have the connection set to restart after 2 seconds not receiving any data (including heartbeats). However, I'm seeing some restarts when I'm pretty sure I still have an internet connection. Thoughts on an appropriate timeout? 2.5 seconds? 3 seconds?

bengotow:
Hey @ericyhwang — looks good! Want to bump the timeout to 5 or 6 seconds? You should receive pings every second, but I imagine the server might not always process the stream events within a second... There won't be any data lost since it'll reconnect with the last cursor it had, so it's just a question of the occasional delay.

We also periodically end the connections server-side, but I think that's after an hour or so.

ericyhwang:
Timeout bumped to 5s.

bengotow:
Awesome - looks good to me! Thanks :-)

ericyhwang:
Thanks! These are all the changes that we (Lever) need to make to the NodeJS library for now. Can you npm publish so we can depend on the new features via package version and not GitHub URL?

bengotow:
Yep—definitely. I'll do it now,

PR #12, opened 2015-07-09
Delta streaming - don't auto-restart when connection is explicitly closed

ericyhwang:
@bengotow 

Bunch of small fixes to delta streaming API:
- Clear timeouts when calling `stream.close()` explicitly so the stream doesn't auto-restart in that case
- Add test for ensuring no auto-restart on explicit close
- Fix the auto-restart test, which failed due to bumping up the timeouts in commit d37708327ebd7e809e063a3f8b34fa7a9df13598

Unrelated, I also added missing err params to callbacks in README.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ var Nylas = require('nylas').config({
 Every resource method accepts an optional callback as the last argument:
 
 ```javascript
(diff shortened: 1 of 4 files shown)


bengotow:
Hey Eric—this looks great. Thanks for adding an additional test!

PR #13, opened 2015-07-10
Add params to event.save method

jylauril:
Added ability to pass in params to the event.save method in order to use them as a query-string value in the url. I followed similar implementation from restful-model-collection with the exception that I made the params optional. This way it also retains the backwards compatibility.

Also added tests for this feature.

### DIFF
--- a/lib/models/event.js
+++ b/lib/models/event.js
@@ -56,13 +56,21 @@
       })
     });
 
-    Event.prototype.save = function(callback) {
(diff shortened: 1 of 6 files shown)


jylauril:
I updated this PR to include the same feature for the model collection's delete method, as the API supports this.

khamidou:
Thanks for the pull request, Jyrki! Merging it.

PR #14, opened 2015-07-11
Fixed calendar model not exposing readOnly state of a calendar

jylauril:
Noticed that the calendar model wasn't exposing the read_only property from the response json. This is very essential to determine if you can create an event to a calendar, so I added it.


khamidou:
Thanks for fixing! Merging.

Issue #1, opened 2015-04-14
Problem with accounts endpoint


misticini:
Hello,

I'm using Nylas-nodejs sdk and i've detected a bug. 

I'm trying to get the accounts with:

```
Nylas.with(ApiToken).accounts.list().then(function(accounts) {
}); 
```
But this functionality have a bunch of bugs. I've patch the NylasSDK but only on Javascript source (not on coffeescript) but i will list the problems that make this not work:

1- 

On management-model-collection.js the code was:
```
ManagementModelCollection.prototype.path = function() {
      return "/a/" + this.api.appId+ "/accounts";
    };
```
This not work because api isnt a defined attribute.

I've patched this to:

```
ManagementModelCollection.prototype.path = function() {
      Nylas = require('../nylas');
      return "/a/" + Nylas.appId+ "/accounts";
    };
```

2- 
On nylas-connection.js the code on line 41 was:
```
if (options.json == null) {
        options.json = true;
      }
 if (this.accessToken) {
        options.auth = {
          'user': this.accessToken,
          'pass': '',
          'sendImmediately': true
        };
}
```
I've patched this  to:

```
if (options.json == null) {
        options.json = true;
      }
      if (options.path.substr(0,3) === '/a/' ) {
        options.auth = {
          'user': Nylas.appSecret,
          'pass': '',
          'sendImmediately': true
        };
      }
      else if (this.accessToken) {
        options.auth = {
          'user': this.accessToken,
          'pass': '',
          'sendImmediately': true
        };
      }
```

I can made a patch in coffeescript source and make a pull request, but i think that with all explained its better to nylas team corrected this bug on your way :)

Regards

bengotow:
Hi @misticini thanks for reporting these issues! I'll take a look. In #1, it looks like the path should reference `this.connection` and not `this.api`. I think in the second one you're totally right—the access token is used for most requests, but the accounts endpoint uses the app secret. I can modify the coffeescript and add test cases for these. Stay tuned!

misticini:
Hello Ben,

it's a pleasure helping this awesome project. I'm using nilas in my MSc Thesis, and maybe i can helping in anything else if i found another bug or if i made any development on Node-js sdk :)

Regards

jstejada:
This should be fixed with https://github.com/nylas/nylas-nodejs/commit/38839d44622879cece6a9e93a47f9639197b8570

PR #6, opened 2015-06-24
Created Expand Method to Support the expanded view of threads

trever:
```js
namespace.threads.expand('c96gge1jo29pl2rebcb7utsbp').then(function(thread) {
    console.log(thread.messages);
}).catch(function(err) {
    console.log('Thread not found! Error: ' + err.toString());
})
```

PR #20, opened 2015-10-09
Clarify and fix account related endpoints

fractallian:
When Nylas.config is called, create a Nylas.accounts endpoint that hits /a/<appid>/accounts and uses a special connection that uses app secret instead of access token. This fetches info about your app's accounts. This endpoint is undocumented?

This also fixes the connection.account endpoint to behave like all others. It fetches info about the connection's account. https://nylas.com/docs/platform/#account

This fixes issue #1


spec/delta-spec.coffee
```diff
@@ -3,6 +3,7 @@ jasmine = require 'jasmine-node'
 {PassThrough} = require 'stream'
 Delta = require '../models/delta'
 NylasConnection = require '../nylas-connection'
+Promise = require 'bluebird'
```
khamidou: Why add this line?

khamidou:
Thanks for fixing this, @fractallian! It makes the code much cleaner.
I only have two remarks I'd like you to address before I merge this pull request:

1. Could you sign our [Contributor License Agreement](https://www.nylas.com/cla.html). It's similar to the ones used by Meteor and NodeJS.
2. I'd prefer that `ManagementAccounts` belonged to a `ManagementModelCollection`, instead of making `RestfulModelCollection` handle both accounts and the other models. Could you change this?

regards,

Karim

spec/delta-spec.coffee
fractallian: I was getting a Promise undefined error when running the test.

fractallian:
You're welcome! Happy to sign the CLA and make your desired change.

fractallian:
On second look, I had intended to remove `ManagementModelCollection` entirely. It is no longer needed after moving `ManagementAccounts` to top level `Nylas` and using a separate connection.

TimFletcher:
I'm struggling to get the account info for a connected account and I believe this fixes it. Could this be merged please? Thanks!

jstejada:
Hey @fractallian, @qbalin, @mccolin, thanks for filing and updating this PR, and sorry for the delay in getting back to you!
It seems that this PR is from a branch that you are separately maintaining and contains commits from PR #6 which has been closed and won't be included (added #40 instead), and other commits for unrelated changes.

It seems that some of these changes are already on master, but it would still be great if we could split the changes here into separate PRs, and I can go ahead and merge them in individually. Otherwise I can cherry pick any of the changes that are still relevant and file them as PRs independently. 

Going to close this one for now. Thanks!

PR #23, opened 2015-12-16
Fix for example app

kaleguy:


khamidou:
Hi @kaleguy,

Thanks for your contribution. It's much appreciated!

Karim

PR #26, opened 2015-12-18
fix to doc link

kaleguy:


khamidou:
Thanks for taking the time to fix this!

PR #27, opened 2016-01-12
Missing a `state` url parameter

slava-tre:
`state` url parametr - optional arbitrary string that is returned as a URL parameter in a redirect_uri

### DIFF
--- a/lib/nylas.js
+++ b/lib/nylas.js
@@ -102,7 +102,10 @@
       if (options.trial == null) {
         options.trial = false;
       }
-      return this.apiServer + "/oauth/authorize?client_id=" + this.appId + "&trial=" + options.trial + "&response_type=code&scope=email&login_hint=" + options.loginHint + "&redirect_uri=" + options.redirectURI;
+      if (options.state == null) {
+        options.state = '';
+      }
+      return this.apiServer + "/oauth/authorize?client_id=" + this.appId + "&trial=" + options.trial + "&response_type=code&scope=email&login_hint=" + options.loginHint + "&redirect_uri=" + options.redirectURI  + '&state=' + options.state;
     };
 
     return Nylas;
--- a/lib/nylas.js
+++ b/lib/nylas.js
@@ -102,10 +102,11 @@
       if (options.trial == null) {
         options.trial = false;
       }
-      if (options.state == null) {
-        options.state = '';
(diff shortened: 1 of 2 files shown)


khamidou:
Thank you for the pull request @slava-tre! It looks good to me, but the SDK is compiled from coffeescript, so you'd need to make the same changes to [nylas.coffee](https://github.com/nylas/nylas-nodejs/blob/master/nylas.coffee#L54)

lib/nylas.js
```diff
@@ -102,7 +102,10 @@
       if (options.trial == null) {
         options.trial = false;
       }
-      return this.apiServer + "/oauth/authorize?client_id=" + this.appId + "&trial=" + options.trial + "&response_type=code&scope=email&login_hint=" + options.loginHint + "&redirect_uri=" + options.redirectURI;
+      if (options.state == null) {
+        options.state = '';
```
khamidou: I'd prefer to not add `state` to the URL if it isn't defined.

slava-tre:
Hi @khamidou! Is my updates - OK?

nylas.coffee
```diff
@@ -51,6 +51,9 @@ class Nylas
     throw new Error("urlForAuthentication() requires options.redirectURI") unless options.redirectURI?
     options.loginHint ?= ''
     options.trial ?= false
-    "#{@apiServer}/oauth/authorize?client_id=#{@appId}&trial=#{options.trial}&response_type=code&scope=email&login_hint=#{options.loginHint}&redirect_uri=#{options.redirectURI}"
+    if options.state?
+      "#{@apiServer}/oauth/authorize?client_id=#{@appId}&trial=#{options.trial}&response_type=code&scope=email&login_hint=#{options.loginHint}&redirect_uri=#{options.redirectURI}&state=#{options.state}"
```
khamidou: I'd prefer not to duplicate the URL string. You should be doing something like this instead:

```coffeescript
url = "#{@apiServer}/oauth/authorize?client_id=#{@appId}&trial=#{options.trial}&response_type=code&scope=email&login_hint=#{options.loginHint}&redirect_uri=#{options.redirectURI}"

if options.state?
  url += "&state=#{options.state}"
```

khamidou:
We're almost there! Fix the comment you've addressed and I'll merge this.

Thanks,

k

slava-tre:
@khamidou Done...

slava-tre:
Ok... @khamidou Could you confirm =)

khamidou:
Thanks for fixing @slava-tre!

PR #30, opened 2016-01-14
Bump delta stream reconnect timeout to 15s, reduce verbosity of reconnect logging

ericyhwang:
@khamidou 

It seems like the Nylas server has recently changed to send heartbeats on the streaming delta connection only every 5 seconds, increased from every 1 second. This client library happens to have been configured to time out and reconnect after 5 seconds with no heartbeat.

Things do still work. However, that's more requests to the Nylas server than needed, plus it results in a lot of log spam.

This PR:
- Increases the default reconnect timeout to 15 seconds, up from 5 seconds. (2-3 missed 5s heartbeats, instead of 4-5 missed 1s heartbeats)
- Puts the reconnect timeout onto `Delta. streamingTimeoutMs` so it can be changed by clients in the future without needing to change this library.
- Reduces internal logging on reconnects to only log the URL, instead of the entire DeltaStream object.

### DIFF
--- a/lib/models/delta.js
+++ b/lib/models/delta.js
@@ -199,7 +199,8 @@
     };
 
     DeltaStream.prototype._restartConnection = function(n) {
-      console.log("Restarting Nylas DeltaStream connection (attempt " + (n + 1) + ")", this);
+      var ref;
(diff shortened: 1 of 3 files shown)


khamidou:
LGTM. Thanks for fixing this!

TimFletcher:
Fantastic! Thanks guys! :smile:

khamidou:
@TimFletcher All credit goes to @ericyhwang !

PR #37, opened 2016-06-22
'version' attribute should be an integer

### LINKED ISSUE #36: Unable to delete drafts
For a delete request, the API requires that the draft version number be included in the request body. As far as I can tell, there's no way to do this yet.

Ideally, this code would work:
`Nylas.with(access_token).drafts.delete(id, {"version": draft.version});`

But this adds the version number as a query string rather than putting it in the request body.

fabiomargarido:
Fixes #36

### DIFF
--- a/models/message.coffee
+++ b/models/message.coffee
@@ -65,7 +65,7 @@ class Message extends RestfulModel
       jsonKey: 'draft'
       queryable: true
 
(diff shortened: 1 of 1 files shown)


fabiomargarido:
Compiled js files are missing

PR #40, opened 2016-08-11
Add option to support view=expanded api param

jstejada:
Currently, for all methods on `RestfulModelCollection` except `find`,
users can manually pass `{view: 'expanded'}` as parameter.
This commit allows request params to be passed to `RestfulModelCollection::find`
and `RestfulModelInstance::get` too, so users can pass any params they
want when fetching individual models.

Additionaly, updates nylas-connection to support the option `expanded: true`
for all requests and transforms it to `view: 'expanded'` for convenience

### DIFF
--- a/models/delta.coffee
+++ b/models/delta.coffee
@@ -2,7 +2,6 @@ backoff = require 'backoff'
 {EventEmitter} = require 'events'
 JSONStream = require 'JSONStream'
 Promise = require 'bluebird'
-querystring = require 'querystring'
 request = require 'request'
 
 module.exports = class Delta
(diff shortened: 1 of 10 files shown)


models/delta.coffee
```diff
@@ -90,15 +92,19 @@ class DeltaStream extends EventEmitter
...
     path = "/delta/streaming"
+    excludeTypes = @params.excludeTypes ? []
+    includeTypes = @params.includeTypes ? []
+
     queryObj =
       cursor: @cursor
-    queryObj.exclude_types = @excludeTypes.join(',') if @excludeTypes?.length > 0
-    queryStr = querystring.stringify(queryObj)
-    path += '?' + queryStr
+    queryObj.exclude_types = excludeTypes.join(',') if excludeTypes.length > 0
```
bengotow: This looks great—my only feedback is that we could make this a bit more concise, since we `? []` and then say `.length > 0`.

Maybe:
```
queryObj.exclude_types = (@params.excludeTypes ? []).join(',')
```

The only thing is that exclude would be '' if the array was zero length.. hmm.

bengotow:
LGTM just one line comment!

khamidou:
LGTM, thanks for fixing. Can you add a changelog entry for your API changes though?

models/delta.coffee
jstejada: Yeah, I think that was intentional, I don't want to add they param unless there is something present

jstejada:
thanks @khamidou , will do

PR #49, opened 2017-06-14
Fix Message.sendRaw

l8on:
Before this change, `Message.sendRaw` would always throw an error after
a successful send, which is the worse time for an email sending client
to respond with an error.

To fix,
* Pass `nylasConnection` as the first parameter to the `Message`
  constructor when creating a message object after a successful send.


l8on:
@pfista @khamidou 
We'd love to stop using our fork if possible.
Any updates here? Has this bug been fixed in a different way and published already?

spang:
Hi @l8on, we're reviewing this right now—thanks for the fix! In the meantime, could you sign our CLA? We need this to merge the pull request. https://docs.google.com/forms/d/e/1FAIpQLSdCnXWsaTkw5jhJ8DMtMu5TUI9Q-El_rCuGzkdBvnNJr8oDWQ/viewform

l8on:
@spang Signed!
I will see about those merge conflicts....

spang:
@l8on Great, thanks!

The bugfix LGTM and I verified that it works, and I have some questions about the additional logging in this patchset. We're hesitant to merge in unsquelchable logging into the SDK because if, for example, someone is building a service that must output JSON logs, this logging won't be customizeable. Generally, we feel that logging should be entirely within the hands of the SDK user, and the SDK by default should not print any output.

Is this logging an essential part of your patch? Are you able to move the logging to the code that calls the SDK, rather than logging within your fork of the SDK?

l8on:
@spang no it is not essential, though it did help diagnose this issue.
I will remove the logging from his patch, I don't think we are relying on those logs for anything critical.

PR #52, opened 2017-06-27
Move .coffee files into src directory

mhahnenberg:
They were littering the top-level directory. This also makes relative
imports work identically to the grunt-generated JS in the lib directory.


src/nylas-connection.coffee
```diff
@@ -22,7 +22,7 @@ Folder = require('./models/folder').Folder
 
 Attributes = require './models/attributes'
 
-PACKAGE_JSON = require './package.json'
+PACKAGE_JSON = require '../package.json'
```
mhahnenberg: Note this change, which fixes the underlying issue when relatively importing package.json

PR #60, opened 2017-07-28
Refactor: Use event to emit DeltaStream retry status

kewitz:
This shouldn't be logging to console if it's an `EventEmitter`.
- [x] removed a redundant `emit(error)`
- [x] added a new event called `info` to deal with the re-connection info

### DIFF
--- a/src/models/delta.coffee
+++ b/src/models/delta.coffee
@@ -57,6 +57,7 @@ Emits the following events:
 - `response` when the connection is established, with one argument, a `http.IncomingMessage`
 - `delta` for each delta received
 - `error` when an error occurs in the connection
+- `info` when the connection status changes
 ###
 class DeltaStream extends EventEmitter
   # Max number of times to retry a connection if we receive no data heartbeats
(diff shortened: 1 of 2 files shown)


src/models/delta.coffee
```diff
@@ -140,11 +141,10 @@ class DeltaStream extends EventEmitter
     @timeoutId = setTimeout @restartBackoff.backoff.bind(@restartBackoff), Delta.streamingTimeoutMs
 
   _onError: (err) ->
-    console.error 'Nylas DeltaStream error:', err
-    @restartBackoff.reset()
     @emit('error', err)
+    @restartBackoff.reset()
 
   _restartConnection: (n) ->
-    console.log "Restarting Nylas DeltaStream connection (attempt #{n + 1}):", @request?.href
```
gregory: some people might be relying on this, i don't think it's a good idea to remove it.

kewitz: Relying on a `console.log`?

gregory: yes - some people might might be monitoring the logs and relying on this to see if anything goes wrong with the delta stream. 
Not sure if this is a blocker though.

kewitz: You're right, that might happen.

gregory: regardless @kewitz you might need to call grunt to compile your changes into the /lib

spang:
Hi @kewitz, thanks for your contribution to the Nylas NodeJS SDK! We're reviewing this now, in the meantime can you sign our CLA so we'll be able to merge the PR?
It's at https://docs.google.com/forms/d/e/1FAIpQLSdCnXWsaTkw5jhJ8DMtMu5TUI9Q-El_rCuGzkdBvnNJr8oDWQ/viewform

spang:
@kewitz I also totally agree that this class shouldn't be logging to the console. Want to add a commit in this PR that :hocho: out the log statements? SDK callers should have what console output is generated entirely within their hands.

I'm ready to merge this as soon as you sign the CLA!

kewitz:
@spang done. Yep, my main objective was to drop all `console.log`s.

PR #63, opened 2017-09-19
add catch statement to range method in restful-model-collection

### LINKED ISSUE #56: unhandled promise rejection - nylas.folders.list() on a gmail account
Hi,

When making a `nylas.folders.list()` call on a GMail account, there's a rejected promise that can't be caught.

I know this call will always fail on GMail accounts, but I'd like to be able to handle the error within my app. Am I missing something? I've tried both `.catch` and callback style. My handlers are never reached.

cdimitroulas:
closes #56 

Errors thrown inside the range method (`RestfulModelCollection.prototype.range`) do not get returned back to the list method.
Added a catch statement to this so that it will work properly.

I am not sure why there are changes to the `lib/models/thread` file (they were made when I ran the default `grunt` command). If this is undesired I can modify the PR so just let me know.

### DIFF
--- a/lib/models/restful-model-collection.js
+++ b/lib/models/restful-model-collection.js
@@ -171,6 +171,8 @@
               accumulated = accumulated.concat(models);
               finished = models.length < REQUEST_CHUNK_SIZE || accumulated.length >= limit;
               return chunkCallback();
+            })["catch"](function(err) {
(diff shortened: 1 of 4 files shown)


spang:
Thank you for your contribution!

PR #67, opened 2017-10-16
passing the offset down from list to range

dejitaiza:
Hey,

As I was trying out some strategies to retrieve the first messages from an account, I noticed the offset wasn't taken into account when calling something like:

```
nylas.with(account.id).messages.list({ limit: 1, offset: 10 });
```

Seems like the offset parameters wasn't being used.

### DIFF
--- a/src/models/restful-model-collection.js
+++ b/src/models/restful-model-collection.js
@@ -55,7 +55,7 @@ export class RestfulModelCollection {
       .request({
         method: 'GET',
         path: this.path(),
-        qs: _.extend({ view: 'count' }, params),
+        qs: _.extend({ view: 'count' }, params)
       })
       .then(function(json) {
         if (callback) {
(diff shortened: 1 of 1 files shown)


jackiehluo:
Cool, that looks good to me! Thanks for the patch! First, do you mind running `yarn prettier` or `npm run prettier` and resubmitting to make sure it passes our (brand new!) linter on Travis?

And second, could you sign our [CLA](https://docs.google.com/forms/d/e/1FAIpQLSdCnXWsaTkw5jhJ8DMtMu5TUI9Q-El_rCuGzkdBvnNJr8oDWQ/viewform) before I merge?

Thanks!

dejitaiza:
Yes sure! I'll look at it this weekend and update my PR accordingly. :-)

jackiehluo:
Hey, @dejitaiza! Just wanted to check back on this patch. Any chance you could update the PR by tomorrow? If not, we might just write our own fix and push it so that we can get it into our next release. We're hoping to do a release in the next day or two!

jackiehluo:
Hey! I just pushed this fix in https://github.com/nylas/nylas-nodejs/commit/903429180b1467a322440e24d8d65ffd2bfc496b. Sorry about that—we want to make sure that it gets into the next release (hopefully tomorrow?), so that's why we didn't have much time to wait. But thank you for taking a look! We really appreciate it! Going to close this PR now.

PR #69, opened 2017-12-14
Fix handling of error when not a JSON request

jylauril:
When sending raw MIME emails, the `json` request option is `false`, which means that the body is still a stringified JSON instead of an object. This means that when ever there was a failure, the error object never included any information about what happened because it was referencing a `.message` property on a string.

This specific request callback might need some more love as it seems to be behaving a little inconsistently. For example, if there is a `body.server_error`, the `error` is suddenly turned into a string instead of an Error instance. Also maybe it would be better to move the `JSON.parse` part on top and figure out what type the body is instead of relying on the request option's `json` value, and then use it in both cases (error & success).

### DIFF
--- a/src/nylas-connection.js
+++ b/src/nylas-connection.js
@@ -103,6 +103,9 @@ module.exports = class NylasConnection {
     return new Promise((resolve, reject) => {
       return request(options, (error, response, body = {}) => {
         if (error || response.statusCode > 299) {
+          if (_.isString(body)) {
+            body = JSON.parse(body);
+          }
           if (!error) {
             error = new Error(body.message);
           }


jackiehluo:
Oh, great catch! Thanks for the PR! Your last point about pulling out the check if we're going to do it in both cases makes sense to me—what about doing that instead? Like:

```
if (options.json === false) {
    body = JSON.parse(body);
}

    ...
} else {
    ....
}
```

In general, checking that `options.json === false` is preferable to checking if the body is a string, since we previously had a bug because the body was a string but wasn't JSON-parsable!

PR #73, opened 2018-02-28
Sample-app patches and debugging improvements

PerStirpes:
@pfista & Nylas team, I was reading over your docs and tinkering with your example apps. This pr fixes a bug that caused an example app to crash and I made a couple of patches. FWIW, I deployed one of your example apps and it's currently being hosted here https://example-tbdurfycwv.now.sh/

* Fix sample-app example from crashing

* Fix unused 'favicon' variable

* Add debug npm scripts for windows and mac

* Add automatic launch of localhost into the browser. Opens localhost in the default browser

* Patch unhandled rejection

* Update docs


annielcook:
@PerStirpes thanks for making these improvements! The gif you added to the readme is quite helpful!

PR #80, opened 2018-08-06
feat(file): add contentDisposition to attributes

AdriVanHoudt:


AdriVanHoudt:
bump

singingwolfboy:
Hi @AdriVanHoudt, sorry for letting this sit for so long! Can you please rebase this pull request, and add a test to cover the new code?

Gerdie:
Hi @AdriVanHoudt thanks for reporting this! The current version of the [Nylas API](https://docs.nylas.com/reference#files) no longer returns this attribute in the json response, only as a header, which we are already handling [here](https://github.com/nylas/nylas-nodejs/blob/master/src/models/file.js#L75). Please feel free to re-open if this changes!

AdriVanHoudt:
Hi @Gerdie 
This seems like an undocumented breaking api change?
This breaks parts of our application :/
Semi related it would be nice to be able to subscribe to the changelog (https://changelog.nylas.com/) to receive these update although I don't think that this change is documented.

Gerdie:
Hi @AdriVanHoudt this was not an API change, but a documentation change to clarify expected behavior. `content_disposition` _is_ returned in the file object from the GET /messages endpoint(s), but it is _not_ returned in the file object in the GET /files endpoint(s). I have filed a feature request to make the attribute accessible to both endpoints, but in the meantime we are updating the documentation to make current behavior explicit. Re-opening this PR given the GET /messages response does surface this data.

AdriVanHoudt:
@Gerdie ah that is great to hear. 👍 for clarifying the docs.
I rebased the branch but I don't have time atm to add tests sorry.

AdriVanHoudt:
Thanks for merging this 🙌

PR #82, opened 2018-09-05
fix: Append server_error to error.message and not reassign variable

### LINKED ISSUE #81: Bug in 4.2.1
```
            if (body.server_error) {
              error = error.message + ' (Server Error: ' + body.server_error + ')';
            }
```
Line 219 in nylas-connection.js

Getting this error (when we receive the error from gmail): ` uncaughtException: Cannot create property 'statusCode' on string 'Daily sending quota exceeded `

Why? Because it assigning a string to `error`

kewitz:
Fixing the error "Cannot create property 'statusCode' on string" when we 
set the `server_error` and we get a `statusCode`.

Fixes #81

### DIFF
--- a/src/nylas-connection.js
+++ b/src/nylas-connection.js
@@ -142,9 +142,6 @@ module.exports = class NylasConnection {
         if (options.json === false) {
           body = JSON.parse(body);
         }
-        if (typeof error === 'string') {
-          error = new Error(error);
-        }
 
         if (error || response.statusCode > 299) {
           if (!error) {
--- a/src/nylas-connection.js
+++ b/src/nylas-connection.js
@@ -142,13 +142,18 @@ module.exports = class NylasConnection {
         if (options.json === false) {
           body = JSON.parse(body);
         }
+        if (typeof error === 'string') {
+          error = new Error(error);
+        }
 
         if (error || response.statusCode > 299) {
           if (!error) {
             error = new Error(body.message);
           }
           if (body.server_error) {
-            error = `${error.message} (Server Error: ${body.server_error})`;
+            error.message = `${error.message} (Server Error: ${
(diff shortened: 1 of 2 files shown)


kewitz:
@annielcook this might interest you.

gunar:
:+1:

gunar:
Actually, this is what you want to do

```js
            if (typeof error === 'string') {
              error = new Error(error)
            }

**Comment by singingwolfboy** [comment]
Hi there, sorry for letting this sit so long! Could you please add an automated test to cover this code?

File: src/nylas-connection.js [hunk 1]
```diff
@@ -142,13 +142,18 @@ module.exports = class NylasConnection {
...
+          error = new Error(error);
+        }
 
         if (error || response.statusCode > 299) {
           if (!error) {
             error = new Error(body.message);
           }
           if (body.server_error) {
-            error = `${error.message} (Server Error: ${body.server_error})`;
+            error.message = `${error.message} (Server Error: ${
```
Gerdie: Good catch!

src/nylas-connection.js
```diff
@@ -142,13 +142,18 @@ module.exports = class NylasConnection {
         if (options.json === false) {
           body = JSON.parse(body);
         }
+        if (typeof error === 'string') {
```
Gerdie: Is this necessary? Under what circumstances would the incoming error be a string, do you have an example?

kewitz: I'm sorry @Gerdie, I don't quite remember why I did this.

Gerdie: Let's remove it for now, and we can revisit it if we get another bug report? Otherwise, lgtm!

PR #89, opened 2019-02-08
Enabling the `tracking` functionality for drafts

philiiiiiipp:
Allow the tracking attribute to be set.

I am not sure if its possible to specify the exact keys inside the object as well, but this at least allows the usage of tracking


src/models/draft.js
```diff
@@ -104,4 +104,7 @@ Draft.attributes = _.extend({}, Message.attributes, {
     modelKey: 'replyToMessageId',
     jsonKey: 'reply_to_message_id',
   }),
+  tracking: _attributes2.default.Object({
```
singingwolfboy: Please refer to `Attributes.Object`, rather than `_attributes2.default.Object`.

philiiiiiipp:
I am really not sure if i rebased that correctly, I have never used it before. I added the tests though and am now using the untranspiled code.

smrutherford:
Thanks for submitting this PR! This functionality was added in #121 .

PR #92, opened 2019-02-18
Standardize on `yarn`, instead of `npm`.

singingwolfboy:


emorikawa:
I'm not sure how opinionated you are about npm vs yarn, but I believe pretty much all of the advantages yarn used to have are now natively incorporated and used in npm. I feel like if we standardize the other way (npm instead of yarn), it'll actually make it a little bit easier for especially new developers to get started. Plus, it's one less brew install and one less dependency people need to know about.

singingwolfboy:
I strongly believe that we should use one or the other, but not both. I don't really care which we use. If you think that `npm` is the better choice, I'll make a new pull request that standardizes on `npm` instead. Sounds good?

emorikawa:
Agreed! Sounds good.

PR #97, opened 2019-02-18
Replace underscore with lodash

singingwolfboy:
[Lodash](https://lodash.com/) is a modern version of [Underscore](https://underscorejs.org/) that is well-integrated into the bundling system. Most notably, it supports importing only the parts of the library that you need, so you don't need to pull in the parts that are unused.

This pull request drops Underscore and replaces it with Lodash.


emorikawa:
LGTM
➕ selective imports.

PR #100, opened 2019-02-20
Remove clone dependency

singingwolfboy:
Lodash has a [`clone`](https://lodash.com/docs/4.17.11#clone) function which does a shallow clone. It also has a [`cloneDeep`](https://lodash.com/docs/4.17.11#cloneDeep) function which does a recursive clone, the way that the [`clone` library](https://github.com/pvorb/clone) does.

I don't think we _need_ to do a deep clone here, which is why I've replaced this dependency with the shallow `clone` from lodash. However, if there's a use-case that requires deep clone, please let me know, and I can use `cloneDeep` from lodash instead.


emorikawa:
👍 LGTM

singingwolfboy:
Yep, just did it for this PR. Now working on the others.

PR #101, opened 2019-02-20
Native promises

singingwolfboy:
The Bluebird GitHub project includes [this note](https://github.com/petkaantonov/bluebird/#note):

> Promises in Node.js 10 are significantly faster than before. Bluebird still includes a lot of features like cancellation, iteration methods and warnings that native promises don't. If you are using Bluebird for performance rather than for those - please consider giving native promises a shot and running the benchmarks yourself.

We don't need the extra features that Bluebird provides, so it's probably better to drop this dependency and go for native promises instead.

### DIFF
--- a/__tests__/file-spec.js
+++ b/__tests__/file-spec.js
@@ -1,4 +1,3 @@
-import Promise from 'bluebird';
 import request from 'request';
 
 import Nylas from '../src/nylas';
@@ -17,7 +16,6 @@ describe('File', () => {
     testContext.file.contentType = 'text/plain';
     testContext.file.filename = 'sample.txt';
     testContext.file.id = 'fileId';
-    return Promise.onPossiblyUnhandledRejection((e, promise) => {});
   });
 
   describe('upload', () => {
@@ -257,7 +255,8 @@ describe('File', () => {
           expect(err).toBe(testContext.error);
           expect(file).toBe(undefined);
           done();
-        });
+        })
+        .catch(() => {});
       });
     });
   });
--- a/__tests__/nylas-api-spec.js
+++ b/__tests__/nylas-api-spec.js
@@ -1,4 +1,3 @@
-import Promise from 'bluebird';
 import request from 'request';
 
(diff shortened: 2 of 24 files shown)


__tests__/file-spec.js
```diff
@@ -257,7 +255,8 @@ describe('File', () => {
           expect(err).toBe(testContext.error);
           expect(file).toBe(undefined);
           done();
-        });
+        })
+        .catch(() => {});
```
grego118: What's the purpose of this empty catch?

__tests__/nylas-api-spec.js
```diff
@@ -83,7 +81,7 @@ describe('Nylas', () => {
 
     test('should return a promise', () => {
       const p = Nylas.exchangeCodeForToken('code-from-server');
-      expect(p instanceof Promise).toBe(true);
+      expect(p).toBeInstanceOf(Promise);
```
grego118: :+1:

singingwolfboy:
Recent versions of Node.js warn about [unhandled promise rejections](https://developers.redhat.com/blog/2016/11/09/node-7-and-promise-rejections-please-handle-them/): promises that reject and don't have a `.catch()` function to handle that rejection. At the moment, these rejections result in a warning, but the warning indicates that in a future version of Node, an unhandled promise rejection could cause the whole process to crash (just like an uncaught exception).

Bluebird provides the `.onPossiblyUnhandledRejection()` event to handle this issue, and I noticed that the existing code was simply swallowing these uncaught rejections. Native promises don't have this event (as far as I know?), so I simply added `.catch()` lines to the promises that didn't have them. Since the existing code was swallowing these rejections, I did the same with the `.catch()` lines I added.

In a future pull request, we might want to examine these rejections more, since they are essentially uncaught exceptions. Some of these tests might be actually broken, but the exception was getting swallowed by the `.onPossiblyUnhandledRejection()` handler. If that's the case, then we should probably fix those tests, but that belongs in a different pull request, not this one.

Does that make sense?

PR #106, opened 2019-03-01
Add X-Nylas-Client-Id header

singingwolfboy:
Python SDK: https://github.com/nylas/nylas-python/pull/112
Ruby SDK: https://github.com/nylas/nylas-ruby/pull/209

### DIFF
--- a/src/nylas-connection.js
+++ b/src/nylas-connection.js
@@ -22,8 +22,9 @@ const SDK_VERSION = PACKAGE_JSON.version;
 const SUPPORTED_API_VERSION = '2.0';
 
 module.exports = class NylasConnection {
-  constructor(accessToken) {
+  constructor(accessToken, { clientId }) {
     this.accessToken = accessToken;
+    this.clientId = clientId;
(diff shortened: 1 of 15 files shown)


src/nylas-connection.js
```diff
@@ -22,8 +22,9 @@ const SDK_VERSION = PACKAGE_JSON.version;
 const SUPPORTED_API_VERSION = '2.0';
 
 module.exports = class NylasConnection {
-  constructor(accessToken) {
+  constructor(accessToken, { clientId }) {
```
grego118: Could we just pass in the client ID as a standalone parameter? What's the purpose of the destructuring?

singingwolfboy: I would prefer if the constructor had only one argument: an object with keys for the different parameters. That's the best way to handle named function arguments in Node, because it's future-proof: your code can only reference the keys that it's interested in.

Unfortunately, we can't change the first argument to an object without breaking backwards compatibility, but we can make sure that any other arguments that need to be added are keys in the object. Does that make sense?

PR #108, opened 2019-03-13
add revoke-all endpoint

smrutherford:
Add support for revoke-all endpoint

### DIFF
--- a/__tests__/management-account-spec.js
+++ b/__tests__/management-account-spec.js
@@ -63,7 +63,7 @@ describe('ManagementAccount', () => {
   });
 
   describe('downgrade', () => {
-    test('should POST to upgrade an account', () => {
+    test('should POST to downgrade an account', () => {
       Nylas.accounts.connection.request = jest.fn(() =>
         Promise.resolve([
           {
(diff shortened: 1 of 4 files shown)


__tests__/management-account-spec.js
```diff
@@ -84,4 +84,48 @@ describe('ManagementAccount', () => {
...
+          {
+            success: 'true',
+          },
+        ])
+      );
+      Nylas.accounts
+        .first()
+        .then(account => account.revokeAll('abc123')) 
+        .then(resp => {
+          expect(Nylas.accounts.connection.request).toHaveBeenCalledWith({
```
singingwolfboy: Please add an assertion that the HTTP request included `keep_access_token=abc123` in the body.

__tests__/management-account-spec.js
```diff
@@ -84,4 +84,48 @@ describe('ManagementAccount', () => {
...
+          {
+            success: 'true',
+          },
+        ])
+      );
+      Nylas.accounts
+        .first()
+        .then(account => account.revokeAll())
+        .then(resp => {
+          expect(Nylas.accounts.connection.request).toHaveBeenCalledWith({
```
singingwolfboy: Please add an assertion that the HTTP request did _not_ include `keep_access_token` in the body.

PR #115, opened 2019-04-02
rm deprecated field in google auth

Gerdie:
Was getting `400: "Error expanding 'fields' parameter. Cannot find matching fields for path 'email'"` responses from google, because the fields param no longer works; dropping it. 

See: https://meta.stackoverflow.com/questions/376116/cannot-sign-in-with-google-error-expanding-fields-parameter-cannot-find-matc


Gerdie:
Update after discussing with @smrutherford --> this field works as a list. Tested & confirmed working. Can't find documentation on this parameter, but _can_ get an illuminating message from misconfiguring field types in https://developers.google.com/apis-explorer/#search/oauth2/oauth2/v2/oauth2.tokeninfo?fields=%255Bemail%255D&_h=6&

PR #121, opened 2019-06-17
Implement message tracking functionality

jieunsharonkim:
Example for README.md:

```  
Nylas.config({
    appId: 'appID',
    appSecret: 'appSecret',
  });

  const nylas = Nylas.with(NYLAS_TOKEN);
  const draft = nylas.drafts.build({
    "tracking":{
    	"opens": true
    },
    "body" : "let's have coffee!",
    "subject": "coffee",
    "to": [
        {
            "name": "test",
            "email": "test@test.com"
        }
    ],
    "from": [
    	{
    		"name": "test2",
    		"email": "test2@test.com"
    }
    ]
});

draft.send().then(message => {
    console.log(`${message.id} was sent`);
  });
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,9 +107,9 @@ To obtain a dynamic list of IP addresses that Nylas might use to connect.
 
 ```javascript
   const nylas = Nylas.with(ACCESS_TOKEN);
-  
+
   // Get IP Addresses
-  
+
   Nylas.accounts.first()
   .then(account => account.ipAddresses())
   .then(response => console.log(response));
@@ -326,6 +326,29 @@ const draft = nylas.drafts.build({
   to: [{ email: 'ben@nylas.com' }],
   replyToMessageId: MESSAGE_ID,
 });
+
+// Enabling tracking
+
+  const draft = nylas.drafts.build({
+    "tracking":{
+    	"opens": true
+    },
+    "body" : "let's have coffee!",
+    "subject": "coffee",
+    "to": [
+        {
+            "name": "test",
+            "email": "test@test.com"
+        }
+    ],
+    "from": [
+    	{
+    		"name": "test2",
+    		"email": "test2@test.com"
+    }
+    ]
+  });
+
 
 // Sending the draft
 
--- a/README.md
+++ b/README.md
@@ -328,9 +328,6 @@ const draft = nylas.drafts.build({
 // Enabling tracking
 
   const draft = nylas.drafts.build({
-    "tracking":{
-    	"opens": true
-    },
     "body" : "let's have coffee!",
     "subject": "coffee",
     "to": [
@@ -350,7 +347,7 @@ const draft = nylas.drafts.build({
 
 // Sending the draft
 
-draft.send().then(message => {
+draft.send(tracking).then(message => {
   console.log(`${message.id} was sent`);
 });
 
--- a/README.md
+++ b/README.md
@@ -325,7 +325,7 @@ const draft = nylas.drafts.build({
   replyToMessageId: MESSAGE_ID,
 });
 
-// Enabling Tracking
+// Enabling tracking
 // NB: When passing in the tracking object, you *must* pass in a value for callback as the first parameter.
 
 const tracking = {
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -107,9 +107,9 @@ To obtain a dynamic list of IP addresses that Nylas might use to connect.
 
 ```javascript
   const nylas = Nylas.with(ACCESS_TOKEN);
-  
+
   // Get IP Addresses
-  
+
```
Gerdie: np: let's remove these newlines added here

src/models/draft.js
```diff
@@ -17,6 +17,8 @@ export default class Draft extends Message {
     if (this.draft) {
       json.object = 'draft';
     }
+
+    json['tracking'] = this.tracking;
```
Gerdie: Thinking about where it makes sense to add this --

`tracking` isn't really an attribute on a Draft object (or a Message object), so defining it inside of `toJSON()` would be a little strange. The `toJSON` method is used in a lot of places like `saveRequestBody()`, which means that changing `toJSON` would have unintended consequences, including sending that `tracking` object to the POST /drafts endpoint when creating a draft, which we don't want to do.

`saveRequestBody()` makes a little more sense, because it's specifically used to generate the body of a request, but since it's used to generate a `save` request (creating the draft), it would be a little weird there, too.

I'm thinking it makes the most sense to have `tracking` (or maybe even `track_opens`, `track_link_clicks`, etc.) be an argument to the `.send()` method, because that method only talks to the /send endpoint, which is the only endpoint that needs/wants tracking data. Does that sound good to you?

smrutherford: I have the exact same thoughts as @Gerdie. Our `tracking` attribute is only available in our API through our sending functionality: https://docs.nylas.com/reference#message-tracking-overview. 
So I agree with @Gerdie on having `tracking` as an argument to the `.send()` method, but I think we can keep it as `tracking` and just have `tracking` be an object with the different types of tracking functionality we have, similar to how it's shown in our docs that I linked above.

README.md
```diff
@@ -327,6 +325,29 @@ const draft = nylas.drafts.build({
   replyToMessageId: MESSAGE_ID,
 });
 
+// Enabling tracking
+
+  const draft = nylas.drafts.build({
+    "tracking":{
```
Gerdie: let's update this to show how we would pass this tracking obj into `send`, instead of passing it into build

__tests__/draft-spec.js
```diff
@@ -44,6 +44,7 @@ describe('Draft', () => {
             headers: undefined,
             reply_to: [],
             reply_to_message_id: undefined,
+            // tracking: {}
```
Gerdie: let's remove this

__tests__/draft-spec.js
```diff
@@ -192,6 +193,7 @@ describe('Draft', () => {
             headers: undefined,
             reply_to: [],
             reply_to_message_id: undefined,
+            tracking: {}
```
Gerdie: this test calls `testContext.draft.save().then(() => ...` and since our new code checks for the existence of `tracking` before adding it to the request body, we won't expect this test to change until we add a tracking argument to the `save` call, like: `testContext.draft.save({"opens": true})`

src/models/draft.js
jieunsharonkim: makes sense, thank you for the feedback !

__tests__/draft-spec.js
Gerdie: now that we've passed a `tracking` object into `send()` as an argument, we should expect to see it in the request body here!

README.md
```diff
@@ -327,9 +325,29 @@ const draft = nylas.drafts.build({
   replyToMessageId: MESSAGE_ID,
 });
 
+// Enabling tracking
+
+  const draft = nylas.drafts.build({
```
Gerdie: np: since we have a code snippet above that calls `const draft = nylas.drafts.build({`let's define a tracking constant here instead of another draft. Then it will be clearer below what `tracking` is when we pass it into `send()`

README.md
```diff
@@ -327,9 +325,19 @@ const draft = nylas.drafts.build({
   replyToMessageId: MESSAGE_ID,
 });
 
+// Enabling tracking
+// NB: When passing in the tracking object, you *must* pass in null as the first parameter.
```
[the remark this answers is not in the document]
jieunsharonkim: great, sounds good!

PR #124, opened 2019-06-28
Increase test coverage for files, contacts, event/email participants

Gerdie:
Brings test coverage from 77.51% to 81.26%

### DIFF
--- a/__tests__/message-spec.js
+++ b/__tests__/message-spec.js
@@ -11,11 +11,12 @@ describe('Message', () => {
   beforeEach(() => {
     testContext = {};
     testContext.connection = new NylasConnection('123', { clientId: 'foo' });
-    testContext.connection.request = jest.fn(() => Promise.resolve());
     testContext.message = new Message(testContext.connection);
     testContext.message.id = '4333';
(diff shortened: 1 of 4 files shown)


__tests__/message-spec.js
```diff
@@ -11,11 +11,14 @@ describe('Message', () => {
...
     testContext.connection = new NylasConnection('123', { clientId: 'foo' });
-    testContext.connection.request = jest.fn(() => Promise.resolve());
     testContext.message = new Message(testContext.connection);
     testContext.message.id = '4333';
     testContext.message.starred = true;
     testContext.message.unread = false;
+    testContext.message.to = [{"email": "foo", "name": "bar"}];
+    testContext.connection.request = jest.fn(() => {
+      return Promise.resolve(testContext.message.toJSON());
+    });
```
smrutherford: nice change! little np: When the only statement in an arrow function is `return`, we can remove `return` and remove the surrounding curly brackets. So the syntax would look like:

```
testContext.connection.request = jest.fn(() => Promise.resolve(testContext.message.toJSON());
```

PR #125, opened 2019-07-01
add billing_state to Account obj

Gerdie:
note: billing_state doesn't exist yet on GET [/account](https://docs.nylas.com/reference#account), so only asserting on GET [.../accounts](https://docs.nylas.com/reference#aclient_idaccounts)


Gerdie:
@grego118 just exposing it for the customer, for them to use if helpful

PR #126, opened 2019-07-02
Contacts

smrutherford:
-Add `groups` and `source` to contact attributes 
-Fix birthday type from date to string
-Update `emailAddresses` attribute in to point to `emails` attribute in API so the SDK will now populate `emailAddresses` attribute
-Update contact test to test for all contact attributes

### DIFF
--- a/__tests__/contact-spec.js
+++ b/__tests__/contact-spec.js
@@ -36,11 +36,12 @@ describe('Contact', () => {
             office_location: undefined,
             notes: undefined,
             picture_url: undefined,
-            email_addresses: [],
+            emails: [],
             im_addresses: [],
             physical_addresses: [],
             phone_numbers: [],
             web_pages: [],
+            groups:[],
           },
           qs: {},
           path: '/contacts',
@@ -69,11 +70,12 @@ describe('Contact', () => {
             office_location: undefined,
             notes: undefined,
             picture_url: undefined,
-            email_addresses: [],
+            emails: [],
             im_addresses: [],
             physical_addresses: [],
             phone_numbers: [],
             web_pages: [],
+            groups:[],
           },
           qs: {},
           path: '/contacts/1257',
@@ -100,13 +102,27 @@ describe('Contact', () => {
       testContext.connection.request = jest.fn(() => {
         const contactJSON = {
           id: '1257',
+          object: 'contact',
+          account_id: '1234',
           given_name: 'John',
           middle_name: 'Jacob',
           surname: 'Jingleheimer Schmidt',
           suffix: 'II',
+          nickname: 'John',
+          birthday: '2019-07-01',
+          company_name: 'Life',
           job_title: 'artist',
-          phone_numbers: [{'type': 'mobile', 'number': '555-444-3333'}],
+          manager_name: 'April',
+          office_location: 'SF, CA, USA',
+          notes: 'lalala',
+          picture_url: 'example.com',
+          emails: [{'type': 'work', 'email': 'john@test.com'}],
+          im_addresses: [{'type': 'yahoo', 'im_address': 'jjj'}],
           physical_addresses: [{'type': 'home', 'city': 'Boston'}],
+          phone_numbers: [{'type': 'mobile', 'number': '555-444-3333'}],
+          web_pages: [{'type': 'blog', 'url': 'johnblogs.com'}],
+          groups: [{id: '123', 'object': 'contact', 'account_id': '1234', 'name': 'Fam', 'path': 'Fam'}],
+          source: 'Contacts',
         };
         return Promise.resolve(contactJSON);
       });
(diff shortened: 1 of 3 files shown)


__tests__/contact-spec.js
```diff
@@ -36,11 +36,12 @@ describe('Contact', () => {
             office_location: undefined,
             notes: undefined,
             picture_url: undefined,
-            email_addresses: [],
+            emails: [],
```
Gerdie: good catch!

__tests__/contact-spec.js
```diff
@@ -100,28 +102,57 @@ describe('Contact', () => {
...
+          groups: [{id: '123', 'object': 'contact', 'account_id': '1234', 'name': 'Fam', 'path': 'Fam'}],
+          source: 'Contacts',
         };
         return Promise.resolve(contactJSON);
       });
     });
 
     test('should resolve with the contact object', done => {
       testContext.contact.save().then(contact => {
+        console.log('THIS IS THE CONTACT', contact);
```
Gerdie: np: rm

__tests__/contact-spec.js
```diff
@@ -100,28 +102,57 @@ describe('Contact', () => {
...
+          manager_name: 'April',
+          office_location: 'SF, CA, USA',
+          notes: 'lalala',
+          picture_url: 'example.com',
+          emails: [{'type': 'work', 'email': 'john@test.com'}],
+          im_addresses: [{'type': 'yahoo', 'im_address': 'jjj'}],
           physical_addresses: [{'type': 'home', 'city': 'Boston'}],
+          phone_numbers: [{'type': 'mobile', 'number': '555-444-3333'}],
+          web_pages: [{'type': 'blog', 'url': 'johnblogs.com'}],
+          groups: [{id: '123', 'object': 'contact', 'account_id': '1234', 'name': 'Fam', 'path': 'Fam'}],
```
Gerdie: np: let's either have strings around every key or no strings around every key (looking at id: '123' vs. 'object': 'contact')

__tests__/contact-spec.js
```diff
@@ -100,28 +102,57 @@ describe('Contact', () => {
...
+          source: 'Contacts',
```
Gerdie: source is either `address_book` or `inbox`

src/models/contact.js
```diff
@@ -10,7 +10,7 @@ class EmailAddress extends RestfulModel {
     return json;
   }
 }
-EmailAddress.collectionName = 'email_addresses';
+EmailAddress.collectionName = 'emails';
```
Gerdie: Does this need to change? What are the implications of changing a `collectionName`?

src/models/contact.js
```diff
@@ -25,7 +25,7 @@ class IMAddress extends RestfulModel {
   toJSON() {
     const json = {
       type: this.type,
-      im_address: this.im_address,
+      im_address: this.imAddress,
```
Gerdie: 💯

src/models/contact.js
```diff
@@ -132,6 +132,39 @@ WebPage.attributes = {
   }),
 };
 
+class Groups extends RestfulModel {
+  toJSON() {
```
Gerdie: np: I think if you define `jsonKey` for all the Groups' attributes below, inheriting the parent RestfulModel's toJSON() will work without defining one here.

src/models/contact.js
```diff
@@ -132,6 +132,39 @@ WebPage.attributes = {
...
+      name: this.name,
+      path: this.path,
+    };
+    return json;
+  }
+}
+Groups.collectionName = 'groups';
+Groups.attributes = {
+  ...RestfulModel.attributes,
+  id: Attributes.String({
```
Gerdie: Group will inherit `id`, `object` and `accountId` from the RestfulModel parent class, so I don't think we need to re-define them here

src/models/contact.js
```diff
@@ -189,7 +222,7 @@ Contact.attributes = {
   }),
   emailAddresses: Attributes.Collection({
     modelKey: 'emailAddresses',
-    jsonKey: 'email_addresses',
+    jsonKey: 'emails',
```
Gerdie: 💯

__tests__/contact-spec.js
```diff
@@ -100,28 +102,57 @@ describe('Contact', () => {
...
+        expect(contact.nickname).toBe('John');
+        expect(contact.birthday).toBe('2019-07-01');
+        expect(contact.companyName).toBe('Life');
         expect(contact.jobTitle).toBe('artist');
-        expect(contact.phoneNumbers[0].toJSON()).toEqual({'type': 'mobile', 'number': '555-444-3333'});
+        expect(contact.managerName).toBe('April');
+        expect(contact.officeLocation).toBe('SF, CA, USA');
+        expect(contact.notes).toBe('lalala');
+        expect(contact.pictureUrl).toBe('example.com');
+        expect(contact.emailAddresses[0].toJSON()).toEqual({'type': 'work', 'email': 'john@test.com'});
```
Gerdie: 💯

__tests__/contact-spec.js
smrutherford: omg lol of course!

__tests__/contact-spec.js
smrutherford: ah, good eye!

__tests__/contact-spec.js
smrutherford: oooh good catch. I updated to `inbox`

src/models/contact.js
smrutherford: ah, good point! I'll remove them

src/models/contact.js
smrutherford: I think the `collectionName` is only referenced in the restful-model file for the `save` and `get` methods in order to make API calls to various endpoints. 
At the end of the day, though, since this collection is an attribute of the Contact model, I don't think it will ever be called and the naming ultimately doesn't matter. All other collection names on the Contact model match our API though so I'm more in favor of changing it. However, I'll change it back to `email_addresses` just in case I'm wrong and it is being used somewhere.

src/models/contact.js
smrutherford: I tried to change this but the tests kept failing. I'm confused on where this `toJSON()` thing is coming from and how it is working so I'm just going to keep it as is. I

smrutherford: Got clarity from Maria and have updated this!

PR #127, opened 2019-07-03
update DELETE qs param to body

Gerdie:
**Docs fix:** The `fs.readFile` parameter `'utf8'` in the readme example for uploading a file was confusing, because [the encoding parameter is optional](https://nodejs.org/api/fs.html#fs_fs_readfile_path_options_callback) and shouldn't be used for many file formats (image files, PDFs, etc.) In the interest of the readme sample code being more copy-pastable, I removed it. 

**Code fix:** Draft deletion was broken, because the draft version (required) was not being passed in the body. The delete method is not easily override-able by the model classes, it's universal to RestfulModelCollection, which I don't love, but changing it seems high impact. Open to better suggestions for how this could work, if you have them. RestfulModelCollection used to allow `/delete` to be called with optional querystring params, but none of the Nylas /delete endpoints take qs params, and only drafts require a request body (trust but verify), so I replaced the querystring params with body.


src/models/restful-model-collection.js
```diff
@@ -181,16 +181,18 @@ export default class RestfulModelCollection {
...
+      body = {};
+    }
+
     const id = itemOrId.id ? itemOrId.id : itemOrId;
+    const deleteBody = itemOrId.id ? itemOrId.deleteRequestBody() : body;
 
-    if (isFunction(params)) {
-      callback = params;
-      params = {};
-    }
```
grego118: Wat?

README.md
```diff
@@ -358,6 +358,12 @@ nylas.drafts
     console.log(`Sent ${message.subject}!`);
   });
 
+// Deleting a draft with the draft object
+nylas.drafts.delete(draft);
+
+// Deleting a draft with the draft id and version
+nylas.drafts.delete('123draftid456', {version: 0});
```
grego118: If `delete` is called from a `drafts` namespace what's stopping us from overriding it? Or is this just one option and it's possible to call `delete` more generically?

src/models/restful-model-collection.js
```diff
@@ -172,7 +172,7 @@ export default class RestfulModelCollection {
     return this._range({ params, offset, limit, path });
   }
 
-  delete(itemOrId, params = {}, callback = null) {
+  delete(itemOrId, body = {}, callback = null) {
```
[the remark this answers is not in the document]
Gerdie: Yeah, working on this :/

src/models/restful-model-collection.js
Gerdie: [discussed in person but commenting for posterity] optional positional args.

smrutherford:
just FYI - DELETE `/labels/{id}`  also requires a request body (`display_name`)

Gerdie:
> just FYI - DELETE `/labels/{id}` also requires a request body (`display_name`)

ah, good catch! addressed that in the latest commit.

__tests__/restful-model-collection-spec.js
```diff
@@ -541,6 +623,78 @@ describe('RestfulModelCollection', () => {
...
+      testContext.collection = new RestfulModelCollection(
+        Draft,
+        testContext.connection
+      );
+      testContext.item = new Draft(testContext.connection, {
+        id: '123', version: 0,
+      });
+    });
+
+    test('should populate body with draft version', () => {
```
smrutherford: similar to above comment. Shouldn't this fail if no draft version is passed in?

src/models/draft.js
```diff
@@ -39,6 +39,16 @@ export default class Draft extends Message {
     return super.saveRequestBody(...arguments);
   }
 
+  deleteRequestBody(params) {
+    var body = {};
+    if (params.hasOwnProperty('version')) {
```
smrutherford: np: let's use ternary statement instead of this if/else statement (i.e. body.version = params.hasOwnProperty('version'))? params.version : this.version)

src/models/restful-model-collection.js
```diff
@@ -181,17 +181,49 @@ export default class RestfulModelCollection {
...
+    return this.deleteItem(options);
+  }
+
+  deleteItem(options) {
+    const item = options.item;
+    const callback = options.callback || null;
+    let body;
+    let qs;
+
+    if (options.hasOwnProperty('body')) {
```
smrutherford: np: let's use ternary statement here too

src/models/restful-model-collection.js
```diff
@@ -181,17 +181,49 @@ export default class RestfulModelCollection {
...
+      body = options.body;
+    } else {
+      body = item.deleteRequestBody({});
+    }
+
+    if (options.hasOwnProperty('qs')) {
```
smrutherford: and ternary statement here too 😅

__tests__/restful-model-collection-spec.js
```diff
@@ -515,6 +538,65 @@ describe('RestfulModelCollection', () => {
...
+        expect(testContext.connection.request).toHaveBeenCalledWith({
+          method: 'DELETE',
+          qs: {},
+          body: {'display_name': 'required'},
+          path: '/labels/123',
+        });
+        done();
+      });
+
+      test('label should populate delete request body if not provided', done => {
```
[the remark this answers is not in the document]
Gerdie: if delete is called with an object that has a displayName, I wanted to populate the request_body with display_name without requiring them to pass it into the method, since it is accessible to the method through the object. Let me try to rename this test to make it more clear what's happening.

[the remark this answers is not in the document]
smrutherford: ahh okay, I thought our API would error out if the displayName wasn't provided but I just checked and it doesn't. 
The test for draft deletion should error out though b/c our docs say:
> Labels can be deleted by issuing a DELETE request to the label’s URI. A label can be deleted even if it still has associated messages.

And I just tried deleting a draft with no request body just to make sure and our API returns the below error:
```
{
    "message": "400 Bad Request: The browser (or proxy) sent a request that this server could not understand.",
    "type": "api_error"
}
```

Gerdie:
Re:
> > just FYI - DELETE `/labels/{id}` also requires a request body (`display_name`) 
> ah, good catch! addressed that in the latest commit.

This was a mistake in the docs. @smrutherford corrected the docs, and I'm removing display_name from the DELETE /labels/id request body.

PR #130, opened 2019-07-19
Switch from babel to tsc, add TypeScript annotations so users get autocomplete

bengotow:
Hey! 👋 I'm working on the Nylas scheduling project and we're building a demo of the new scheduling features in NodeJS. I was looking for typescript definitions for the SDK but didn't find any, so I thought it might be nice to add type annotations to this module!

Here's what I changed:

- `npm build` runs `tsc` (the typescript compiler) instead of `babel`. Thankfully this project was only using babel with a few transforms, and they're also supported in TypeScript.

- The file extensions were switched to `.ts` and I went through and added type annotations everywhere the compiler required. I only made a few meaningful changes to the code—this entire diff is really just annotations.

- When you run `npm build`, it generates the same `lib` folder as before, but with `.d.ts` files as well as the `.js` files. The package.json specifies a root `types` file, and when you use `nylas-nodejs` in your project, VSCode and Atom will auto-detect the types and do super cool stuff. (Screenshots below!)

Screen Shot 2019-07-19 at 2 54 20 AM
Screen Shot 2019-07-19 at 2 53 03 AM
Screen Shot 2019-07-19 at 2 51 51 AM

I found a few small bugs while doing this! These are fixed in this PR (and are I think the only code changes):

- `RestfulModel._get` does not use the callback passed to it: https://github.com/nylas/nylas-nodejs/blob/master/src/models/restful-model.js#L91. (This is used in the `Contact.getPicture` method)

- Attributes.Object does not actually support an `itemClass` param, so `thread.folder` is currently a plain JS object and not a Folder object. https://github.com/nylas/nylas-nodejs/blob/master/src/models/thread.js#L79

The tests still pass and super wonderful and thorough! I was really excited to see that there were so many tests for this, since it makes verifying that the behavior didn't change much easier. I had to move the tests into the `src` folder so they can be passed through the `tsc` compiler as well, but that was the only change there.

---- 

Sidenote: I didn't notice until just now that there's work being done on this repo every week! I know this is a big change, feel free to reject it if it conflicts with other stuff being done. Since the module doesn't change much, we could copy the `.d.ts` files from this branch into master and ship them as static files (rather than generating them from source each build). They might fall out of date eventually, but it'd still be really helpful.


Gerdie:
Hey @bengotow thanks for this! We're going to try to merge some existing work before we circle back to this, but it's on our to-do list!

al1271:
@Gerdie Think you will be able to circle back around to this soon?

Gerdie:
> @Gerdie Think you will be able to circle back around to this soon?

@al1271 yes this is being prioritized

Gerdie:
@bengotow closing this for https://github.com/nylas/nylas-nodejs/pull/173 thanks!

PR #136, opened 2019-07-31
rm ruby SDK refence - we don't use it in example code

Gerdie:


example/authentication/readme.md
```diff
@@ -6,7 +6,7 @@ APIs](https://www.nylas.com/docs/platform#native_authentication).  You can
 connect both Exchange accounts and Gmail account's in this example.
 
 It shows how to receive a `refresh_token` from Google before authenticating with
-Nylas. Then it uses the Nylas Ruby SDK to connect an email account and load the
+Nylas. Then it uses the Nylas Node SDK to connect an email account and load the
```
smrutherford: np: Is there a reason there's a + before Nylas? Can we remove?

Gerdie: that's just how github displays what has been added in the PR, the '+' isn't actually in the file

PR #138, opened 2019-08-15
implement webhooks CRUD

Gerdie:
Adds create, read, update and delete functionality for webhooks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -371,6 +371,31 @@ nylas.messages.search('Hey!').then(messages => console.log(messages));
 
 ```
 
+Creating and Updating Webhooks
+-----
+You can programmatically create, read, update and delete webhooks.
+
+```javascript
+// create a new webhook
+const newWebhook = Nylas.webhooks.build({
+  callbackUrl: 'https://wwww.myapp.com/webhook',
+  state: 'active',
+  triggers: ['event.created', 'event.updated'],
+  });
+newWebhook.save().then(webhook => console.log(webhook.id));
+
+// get and update the state of an existing webhook
+// note: a webhook's callbackUrl & triggers are not updateable, only its state.
+Nylas.webhooks.find('existingWebhookId').then(existingWebhook => {
+  existingWebhook.state = 'active';
+  existingWebhook.save();
+})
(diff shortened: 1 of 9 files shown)


src/nylas.js
```diff
@@ -29,16 +30,15 @@ class Nylas {
       this.apiServer = apiServer;
     }
 
-    let conn;
+    const conn = new NylasConnection(this.appSecret, { clientId: this.appId });
+    this.webhooks = new ManagementModelCollection(Webhook, conn, this.appId);
```
smrutherford: oooh smart - i like how you made it accessible on the Nylas model so ppl don't have to make their own instance of an account management model

src/models/restful-model.js
```diff
@@ -46,6 +46,15 @@ export default class RestfulModel {
     return json;
   }
 
+  // Subclasses should override this method.
+  pathPrefix() {
+    return '';
+  }
+
+  saveEndpoint() {
+    return `${this.pathPrefix()}/${this.constructor.collectionName}`;
```
smrutherford: crafty! 💯

src/models/restful-model-collection.js
```diff
@@ -208,7 +208,7 @@ export default class RestfulModelCollection {
   }
 
   build(args) {
-    const model = new this.modelClass(this.connection);
+    const model = this._createModel({});
```
smrutherford: 👍

README.md
```diff
@@ -371,6 +371,30 @@ nylas.messages.search('Hey!').then(messages => console.log(messages));
...
+```javascript
+// create a new webhook
+const newWebhook = Nylas.webhooks.build({
+  callbackUrl: 'https://wwww.myapp.com/webhook',
+  state: 'active',
+  triggers: ['event.created', 'event.updated'],
+  });
+newWebhook.save().then(webhook => console.log(webhook.id));
+
+// get and update an existing webhook
```
yusra-ahmed: the readme might also be a good place to explain that you can only update state, if they want to update anything else they'll have to make a new post.

PR #141, opened 2019-08-23
add Connect class skeleton for native auth

Gerdie:


src/nylas.js
```diff
@@ -64,6 +65,7 @@ class Nylas {
     const conn = new NylasConnection(this.clientSecret, {
       clientId: this.clientId,
     });
+    this.connect = new Connect(conn);
```
Gerdie: @smrutherford I ended up making this lowercase-c connect because everything else on Nylas (accounts, connection, webhooks) is lowercase. thoughts?

smrutherford: nope - makes sense to me!

PR #142, opened 2019-08-26
implement /connect/authorize

smrutherford:
I initially was going to have a test that tested compatibility with every provider, which is why there are so many email credential variables. But then didn't really see the benefit of it - what are your thoughts?

Also, while implementing this, I didn't see the point of the if statement you put in the Connect class constructor. Since the connection is instantiated in `nylas.js` from NylasConnection, won't the connection always be an instance of NylasConnection?

### DIFF
--- a/__tests__/connect-spec.js
+++ b/__tests__/connect-spec.js
@@ -1,0 +1,117 @@
+import Nylas from '../src/nylas';
+
+describe('Connect', () => {
+
+  const name = 'Connect Test';
+  const password = 'connecting123';
+  const scopes = 'email, calendar, contacts';
+  const exchangeEmail = 'connect_test@outlook.com';
+  const gmailEmail = 'connect_test@gmail.com';
+  const imapEmail = 'connect_test@yahoo.com';
+  const o365Email = 'connect_test@nylatest.onmicrosoft.com';
+  const CLIENT_ID = 'abc';
+  const authorizeParams = ((email, provider, settings, reauthAccountId) => {
+  	return {
+	  	name: 'Connect Test',
+	  	emailAddress: email,
+	  	provider: provider,
+	  	settings: settings,
+	  	scopes: scopes,
+	  	reauthAccountId: reauthAccountId,	
+  	}
+  });
+  const authorizeJSON = {
+  	code: 'bOjq4Wt9ZAlCy0CSbVeDGDQ5PquytC',
+  }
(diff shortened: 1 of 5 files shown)


src/nylas-connection.js
```diff
@@ -147,7 +147,7 @@ module.exports = class NylasConnection {
 
         if (error || response.statusCode > 299) {
           if (!error) {
-            error = new Error(body.message);
+            error = new Error(JSON.stringify(body));
```
smrutherford: I updated this because the auth error code spits out the following response:

```
{
             "message": "Please include the missing fields",
             "type": "auth_error",
             "missing_fields": [
                 "provider"
             ],
             "error": "Missing fields."
}
```
If we only surface the error response's `body.message` value then it masks the error and isn't as helpful.
Not sure if it would have any negative affects on other endpoints...at the top of my head I don't see how it would, but would love to hear your thoughts!

__tests__/connect-spec.js
```diff
@@ -0,0 +1,70 @@
+import Nylas from '../src/nylas';
+
+describe('Connect', () => {
+
+  const name = 'Connect Test';
+  const password = 'connecting123';
+  const scopes = 'email, calendar, contacts';
```
Gerdie: np: let's rm the spaces and use granular scopes, like 'email.modify,email.send'

__tests__/connect-spec.js
```diff
@@ -0,0 +1,70 @@
...
+  const exchangeEmail = 'connect_test@outlook.com';
+  const gmailEmail = 'connect_test@gmail.com';
+  const imapEmail = 'connect_test@yahoo.com';
+  const o365Email = 'connect_test@nylatest.onmicrosoft.com';
+  const CLIENT_ID = 'abc';
+  const authorizeOptions = ((email, provider, settings) => {
+  	return {
+	  	name: 'Connect Test',
```
Gerdie: np: can we use `name` here?

src/nylas-connection.js
```diff
@@ -147,7 +147,7 @@ module.exports = class NylasConnection {
...
```
Gerdie: hmm. I do definitely want to capture the rest of the data in this error message but I want to format it nicely. Maybe for now can we just add another `if` clause below to format this particular edge case, and let's revisit Error handling later?

Issue #146, opened 2019-09-06
Creating & Updating events only supports start_time and end_time


smrutherford:
Needs to support the other available options for the `when` attribute:

-time
-date
-start_date and end_date

jasonbreault:
As expected with the creating a new event, (by passing in the "when" object), when updating an existing event, the event.save() function should support update the date/time of the object by passing the following when object (as stated by Nylas docs):

   when: {
      start_time: <unix time format>,
      end_time: <unix time format>
   }

   when: {
      start_date: <YYYY-MM-DD format>,
      end_date: < YYYY-MM-DD format>
   }
 
when: {
   date: <YYYY-MM-DD format>
   }

smrutherford:
Hi @jasonbreault, thanks again for reporting this issue. We've released a fix and it is now available in the latest NodeSDK v4.8.0. Now, you may create or update events by either assigning the `start` and `end` params or by assigning the `when` parameter.

PR #149, opened 2019-09-17
support time, date, start_date, and end_date for creating and updating events

smrutherford:
This is to address issue #146 

I'm unsure how we should deal with the Event attributes b/c it seems as though only one attribute type can be associated to a param. Also, I don't think we can support updating and creating events using the `when` param as well as the `start` and `end` params b/c then we wouldn't know which param to use to create/update the event. Open to suggestions or ideas that I might've overlooked though!

### DIFF
--- a/__tests__/event-spec.js
+++ b/__tests__/event-spec.js
@@ -94,6 +94,7 @@ describe('Event', () => {
     });
 
     test('should create event with time when start and end are the same UNIX timestamp', done => {
+      testContext.event.when = {};
       testContext.event.start = 1408875644;
       testContext.event.end = 1408875644;
       testContext.event.save().then(() => {
@@ -125,6 +126,7 @@ describe('Event', () => {
     });
 
     test('should create event with start_time and end_time when start and end are different UNIX timestamps', done => {
+      testContext.event.when = {};
       testContext.event.start = 1409594400
       testContext.event.end = 1409598000
       testContext.event.save().then(() => {
@@ -157,6 +159,7 @@ describe('Event', () => {
     });
 
     test('should create event with date when start and end are same ISO date', done => {
+      testContext.event.when = {};
       testContext.event.start = '1912-06-23';
       testContext.event.end = '1912-06-23';
       testContext.event.save().then(() => {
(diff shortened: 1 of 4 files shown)


__tests__/event-spec.js
```diff
@@ -93,13 +93,267 @@ describe('Event', () => {
...
+
     describe('when the request succeeds', () => {
       beforeEach(() => {
         testContext.connection.request = jest.fn(() => {
           const eventJSON = {
             id: 'id-1234',
             title: 'test event',
-            when: {},
+            start: 1409594400,
+            end: 1409594400,
```
Gerdie: our API doesn't return start & end, right? if not, let's just include `when` in this example JSON response

src/models/event.js
```diff
@@ -3,14 +3,67 @@ import Attributes from './attributes';
...
   saveRequestBody() {
     const dct = this.toJSON();
-    if (this.start && this.end) {
-      dct['when'] = {
-        start_time: this.start.toString(),
-        end_time: this.end.toString(),
-      };
-    }
     delete dct['_start'];
     delete dct['_end'];
```
Gerdie: let's rm these too, i dont think they'll be there anymore

src/models/event.js
```diff
@@ -3,14 +3,67 @@ import Attributes from './attributes';
...
+      this.when.date;
+  }
+
+  set start(val) {
+    if (this.when) {
+      if (typeof val === 'number') {
+        if (val === this.when.end_time) {
+          this.when = { time: val }
+        } else {
+          delete this.when.time;
```
Gerdie: should we also proactively delete `when.start_date` and `when.date`, just in case?

src/models/event.js
```diff
@@ -3,14 +3,67 @@ import Attributes from './attributes';
...
+          this.when.start_time = val;        
+        }   
+      }
+      if (typeof val === 'string') {
+        if (val === this.when.end_date) {
+          this.when = { date: val }
+        } else {
+          delete this.when.date;
```
Gerdie: same q

src/models/event.js
```diff
@@ -3,14 +3,67 @@ import Attributes from './attributes';
...
+    } 
+  }
+
+  set end(val) {
+    if (this.when) {
+      if (typeof val === 'number') {
+        if (val === this.when.start_time) {
+          this.when = { time: val }
+        } else {
+          delete this.when.time;
```
Gerdie: same q

src/models/event.js
```diff
@@ -3,14 +3,67 @@ import Attributes from './attributes';
...
+          this.when.end_time = val;        
+        }   
+      }
+      if (typeof val === 'string') {
+        if (val === this.when.start_date) {
+          this.when = { date: val }
+        } else {
+          delete this.when.date;
```
Gerdie: same q

__tests__/event-spec.js
```diff
@@ -93,13 +93,267 @@ describe('Event', () => {
...
+            when: { time: 1409594400 },
```
Gerdie: np: and i think the api returns `{ time: 1409594400, object: "time" }`

smrutherford: true - but in the node SDK, we delete the object attribute in fromJSON() so i'm going to leave this as is

__tests__/event-spec.js
smrutherford: you're right! i'll remove start and end

src/models/event.js
smrutherford: good point

src/models/event.js
smrutherford: good point! i'm going to delete the entire saveRequestBody() method on the event class then since there's nothing unique that needs to be added and the event class inherits saveRequestBody() from the restful model.

PR #156, opened 2019-10-10
implement connect.gmailAccount

Gerdie:
A gmail-specific convenience wrapper for native authentication.

### DIFF
--- a/src/models/connect.js
+++ b/src/models/connect.js
@@ -1,8 +1,10 @@
 export default class Connect {
-  constructor(connection, clientId, clientSecret) {
+  constructor(connection, clientId, clientSecret, googleClientId, googleClientSecret) {
     this.connection = connection;
     this.clientId = clientId;
     this.clientSecret = clientSecret;
+    this.googleClientId = googleClientId;
+    this.googleClientSecret = googleClientSecret;
   }
 
   authorize(options = {}) {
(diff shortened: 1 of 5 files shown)


src/models/connect.js
```diff
@@ -58,4 +60,24 @@ export default class Connect {
   newAccount() {
     // this.authorize() -> this.token()
   }
+
+  async gmailAccount({ name, emailAddress, scopes, googleClientId, googleClientSecret, googleRefreshToken }) {
```
smrutherford: so I also tried to do an async function for `newAccount` and I kept getting this error when compiling: `Uncaught ReferenceError: regeneratorRuntime is not defined` . From this github issue, looks like it is solvable by making changes to the babel config file: https://github.com/babel/babel/issues/7824 . Did yours compile successfully as is though? I wonder if I did something wrong 🤔

src/models/connect.js
```diff
@@ -58,4 +60,24 @@ export default class Connect {
...
+        googleClientSecret: googleClientSecret || this.googleClientSecret,
+        googleRefreshToken: googleRefreshToken
+      }
+    }
+
+    if (scopes) {
+      googleOptions.scopes = scopes;
+    }
+
+    token = await this.authorize(googleOptions);
```
smrutherford: if we got the async/await route, will errors from authorize() still be surfaced? 
also, let's assign token as a const (const token = await this.authorize(googleOptions))

Gerdie:
closing to iterate & add tests

Issue #174, opened 2020-03-28
Replace deprecated "request"


TrySound:
"request" package is deprecated.

https://github.com/request/request#deprecated

A few possible alternatives
- https://github.com/node-fetch/node-fetch
- https://github.com/axios/axios
- https://github.com/sindresorhus/got

Gerdie:
Thanks for reporting, @TrySound

TrySound:
FYI I'm working on this.

mrashed-dev:
`node-fetch` has completely replaced the `request` library, will be available in next release.

Issue #172, opened 2020-03-25
Nylas client crashing node on invalid requests


Pagebakers:
Nylas intermittendly makes our nodes apps crash, even though we have proper error handling.
There seems to be an issue in the client when requests timeout or fail.

Here's a excerpt of the error

"at Request._callback ([..]/node_modules/nylas/lib/nylas-connection.js:177:37)"
TypeError: Cannot read property 'headers' of undefined" 

I'm not sure how to reproduce this and which request is causing this.

The error is happening both on our webhook handers and when sending messages through the api.

Gerdie:
Thanks for reporting @Pagebakers we'll look into this.

Pagebakers:
Any updates on this? this is a real serious issue, it's causing quite a lot of crashes.

Pagebakers:
To help you along a bit...

https://github.com/nylas/nylas-nodejs/blob/master/src/nylas-connection.js#L133

The response needs to be checked, as it can be undefined

Gerdie:
Thanks @Pagebakers feel free to review https://github.com/nylas/nylas-nodejs/pull/177 and let us know if more information in the error message would be useful.

Gerdie:
This was resolved and released yesterday in v4.10.1

PR #177, opened 2020-05-06
check for undefined response and reject

Gerdie:
Re: https://github.com/nylas/nylas-nodejs/issues/172 users reporting bug where application crashes due to an undefined response. 

Taking this as a another reason to prioritize moving off of request, re: https://github.com/nylas/nylas-nodejs/issues/174

# License

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### Unreleased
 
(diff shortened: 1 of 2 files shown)


Pagebakers:
Should be good, although `error` could already be set with an error message. For example response time out or TLS error.

```
error = error || new Error('No response')
```

PR #186, opened 2020-06-04
Fix empty body due to re-initilization

jhatch28:
In the generated (not sure the ts/restfulmodel lingo) is set by default to an empty string,  it's causing the body to be cleared out. 
I'm not sure this fix is what will solve it, but I edited the generated javascript file in my local `node_modules` to not initialize the body to an empty string and it worked. I assume this is where that generated comes from, but I also don't know how to compile typescript into js.

Here's the code in the generated file I modified, `lib/models/message.js` (not `message.d.ts`)
```
var Message = /** @class */ (function (_super) {
    __extends(Message, _super);
    function Message() {
        var _this = _super !== null && _super.apply(this, arguments) || this;
        _this.subject = '';
        _this.to = [];
        _this.cc = [];
        _this.bcc = [];
        _this.body = '';         // Removing the `= '';` part fixed the bug
        return _this;
    }
```

Here's the code that wasn't working before I made the above change:
```
nylas.messages.list({limit: 10}).then(messages => {
    messages.forEach((msg) => console.log(msg.body))
});
```
^ Logs 10 blank lines without change, then full bodies with change.

# License


jhatch28:
Thanks for the test fixes! Ran it locally to double check it works.

PR #190, opened 2020-06-30
support get resources

Gerdie:
Re-opening https://github.com/nylas/nylas-nodejs/pull/189 against `main`

# License


__tests__/resource-spec.js
```diff
@@ -0,0 +1,60 @@
+import request from 'request';
```
TrySound, line 1: 😢

PR #191, opened 2020-07-06
Calendar CRUD

Gerdie:
- Supporting POST & PUT for calendars. 
    - Adds isPrimary, location and timezone attributes.
    - Expand test coverage
- Supporting DELETE for calendars
    - Refactor delete message to return JSON response (previous delete responses were empty, so this was not needed)
    - Expand test coverage

Will address job statuses in a follow-up PR

# License

### DIFF
--- a/__tests__/calendar-spec.js
+++ b/__tests__/calendar-spec.js
@@ -1,0 +1,55 @@
+import request from 'request';
+
+import Nylas from '../src/nylas';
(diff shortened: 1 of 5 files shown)


__tests__/calendar-spec.js
```diff
@@ -0,0 +1,106 @@
+import request from 'request';
```
TrySound, line 1: 😢

PR #195, opened 2020-07-26
Add object attribute back to event.when

Gerdie:
# License

### DIFF
--- a/src/models/event.ts
+++ b/src/models/event.ts
@@ -108,15 +108,6 @@ export default class Event extends RestfulModel {
     return this._save(params, callback);
   }
 
-  fromJSON(json: { [key: string]: any }) {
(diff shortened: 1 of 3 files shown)


src/models/event.ts
```diff
@@ -108,15 +108,6 @@ export default class Event extends RestfulModel {
     return this._save(params, callback);
   }
 
-  fromJSON(json: { [key: string]: any }) {
-    super.fromJSON(json);
-
-    if (this.when) {
-      delete this.when.object;
```
Gerdie, line 115: I'm not sure why this was here, the git history is a migration from coffeescript. The way we've handled event.start and event.when has changed a lot since then, so it's safe to say we don't need this anymore.

PR #197, opened 2020-07-28
Fix the file_ids array reading of toJSON method

Guiradi:
While trying to send the fileIds array in an draft instance, I've catched the error: error:  "TypeError: this.fileIds is not a function" on Draft.toJSON method. 
The error is solved by taking the fileIds parameter on the method and not the fileIds() (function).

# License

### DIFF
--- a/src/models/draft.ts
+++ b/src/models/draft.ts
@@ -14,7 +14,7 @@ export default class Draft extends Message {
       throw Error('toJSON() cannot be called for raw MIME drafts');
     }
     const json = super.toJSON();
-    json.file_ids = this.fileIds();
+    json.file_ids = this.fileIds;
     json.object = 'draft';
 
     return json;


Gerdie:
Hi @Guiradi can you share the code that's throwing that error? [fileIds is a function](https://github.com/nylas/nylas-nodejs/blob/main/src/models/message.ts#L51) so we don't want to remove the function call --> is it possible your code is overwriting fileIds on the draft object?

Guiradi:
Yes, my code is actually overwriting fileIds on the draft object, just like the documentation:

image

Here's my code:

image

image

And the attachments console:

image

image

And the error:

image

But when I change the code based on what fileIds function do:

image

I get success!

So, in fact, the documentation is wrong, even in the [body params explication from the draft post](https://docs.nylas.com/reference#post-draft)

image

Gerdie:
Ah, thanks for reporting. I'll make a ticket to update the documentation.

PR #200, opened 2020-09-08
Be consistent about callback parameters

Gerdie:
We should consistently be passing callback as the last argument to public functions. Added tests for backwards compatibility, so users can still use the old parameter ordering if they want to.

# License


lkaileh (approved):
I ran the tests locally and they all passed! 🎉

PR #204, opened 2020-09-25
Avoid commonjs and circular dependencies

TrySound:
"require" was used in nylas connection to get options from static
class properties.

In this diff I moved the source of data to separate module which can be
imported by any module and refactored static properties with getters and
setters to preserve existing behaviour.

This is necessary simplification before replacing "request" with `node-fetch`.

# License

### DIFF
--- a/src/nylas.ts
+++ b/src/nylas.ts
@@ -1,5 +1,5 @@
 import request from 'request';
-
+import * as config from './config';
 import NylasConnection from './nylas-connection';
 import ManagementAccount from './models/management-account';
 import Account from './models/account';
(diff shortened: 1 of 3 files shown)


src/nylas.ts
```diff
@@ -10,24 +10,32 @@ import Webhook from './models/webhook';
...
     clientSecret,
     apiServer,
-    appId,
-    appSecret
-  } : {
-    clientId: string,
-    clientSecret: string,
-    apiServer?:string,
-    appId?: string,
-    appSecret?: string
```
TrySound, line 30: appId and appSecret looks like unused

TrySound:
cc @Gerdie

TrySound:
Friendly ping @Gerdie

sdwvit (approved):
Lgtm

PR #210, opened 2020-11-21
Make connection.request types more strict

TrySound:
Currently types are specified as weak type `{[string]: any }`. Such type
make it quite hard to replace deprecated "request" with something else
(I'm planning to use node-fetch).  In this diff I specified all fields
explicitly and made `path` a mandatory field.

# License

### DIFF
--- a/src/nylas-connection.ts
+++ b/src/nylas-connection.ts
@@ -25,6 +25,17 @@ const PACKAGE_JSON = require('../package.json');
 const SDK_VERSION = PACKAGE_JSON.version;
 const SUPPORTED_API_VERSION = '2.1';
 
+export type RequestOptions = {
+  path: string;
+  method?: string;
+  headers?: { [key: string]: string };
+  qs?: { [key: string]: any };
+  downloadRequest?: boolean;
+  json?: boolean;
+  formData?: any;
+  body?: any;
+};
+
 export default class NylasConnection {
   accessToken: string | null | undefined;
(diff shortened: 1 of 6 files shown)


TrySound:
cc @sdwvit

src/nylas-connection.ts
```diff
@@ -51,66 +62,60 @@ export default class NylasConnection {
...
-      }
-    }
-    if (options.json == null) {
-      options.json = true;
+  requestOptions(options: RequestOptions) {
+    const newOptions: CoreOptions & UrlOptions = {
+      method: options.method || 'GET',
+      url: `${config.apiServer}${options.path}`,
+      headers: options.headers || {},
+      json: options.json == null ? true : options.json,
```
sdwvit, line 70:
```suggestion
      json: options.json || true,
```
? don't know for sure if json=undefined (or any falsy value) makes any sense here

sdwvit (approved):

src/nylas-connection.ts
TrySound: This will not let to set `json: false`

TrySound: true is default == null checks if it's passed

TrySound:
Friendly ping @sdwvit @Gerdie

src/nylas-connection.ts
sdwvit: 👍

PR #214, opened 2021-01-21
Migrate exchangeCodeForToken to node-fetch

TrySound:
Ref https://github.com/nylas/nylas-nodejs/issues/174

Migrating whole library from "request" at once would be quite massive.
In this diff I migrated only exchangeCodeForToken to figure out how to
test and install dependencies.

# License

### DIFF
--- a/src/nylas.ts
+++ b/src/nylas.ts
@@ -1,4 +1,6 @@
-import request from 'request';
+// TODO since node 10 URL is global
+import { URL } from 'url';
+import fetch from 'node-fetch';
 import * as config from './config';
 import NylasConnection, { RequestOptions } from './nylas-connection';
 import ManagementAccount from './models/management-account';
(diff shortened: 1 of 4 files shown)


TrySound:
cc @sdwvit

src/nylas.ts
```diff
@@ -130,38 +132,36 @@ class Nylas {
...
-      return request(options, (error, response, body) => {
-        if (!body || !body['access_token']) {
-          let errorMessage = 'No access token in response';
-          if (body && body.message) errorMessage = body.message;
-          error = error ? error : new Error(errorMessage);
+    const url = new URL(`${this.apiServer}/oauth/token`);
+    url.searchParams.set('client_id', this.clientId);
+    url.searchParams.set('client_secret', this.clientSecret);
+    url.searchParams.set('grant_type', 'authorization_code');
+    url.searchParams.set('code', code);
```
sdwvit, line 139: using URL is an interesting choice here ;) is it just to urlencode querystring? looks good to me, just wanted to point out

src/nylas.ts
```diff
@@ -130,38 +132,36 @@ class Nylas {
...
-        } else {
-          resolve(body['access_token']);
+          return body['access_token'];
+        },
+        error => {
           if (callback) {
-            return callback(null, body['access_token']);
+            callback(error);
           }
+          throw error;
```
sdwvit, line 162: wouldn't it make sense to recreate an Error so it includes current frame in a stacktrace?

src/nylas.ts
TrySound: Browser support makes it more preferred choice over `querystring`

src/nylas.ts
TrySound: Done

TrySound:
Hi @sdwvit @Gerdie, is it good to go?

PR #219, opened 2021-03-24
Pass error message

Tranced:
Just wanted to pass along the error message generically in our SDK if customers come upon them.
Looks like we only pass status code.

For context, a customer is using native auth and is only seeing a 401.
However, in our authentication example(which just uses a regular request/not the SDK), we're able to see the full stack trace.

Before, the message from the SDK was:

```
Error:
    at Request._callback (/home/oleg/Documents/bs/address-book/node_modules/nylas/lib/nylas-connection.js:140:33)
    at Request.self.callback (/home/oleg/Documents/bs/address-book/node_modules/request/request.js:185:22)
    at Request.emit (events.js:314:20)
    at Request.<anonymous> (/home/oleg/Documents/bs/address-book/node_modules/request/request.js:1154:10)
    at Request.emit (events.js:314:20)
    at IncomingMessage.<anonymous> (/home/oleg/Documents/bs/address-book/node_modules/request/request.js:1076:12)
    at Object.onceWrapper (events.js:420:28)
    at IncomingMessage.emit (events.js:326:22)
    at endReadableNT (_stream_readable.js:1241:12)
    at processTicksAndRejections (internal/process/task_queues.js:84:21) {
  statusCode: 401
}
```

now it's:

```
Error: Nylas_Error: You've entered invalid credentials.  Provider_Error: Wrong username or password for https://outlook.com/EWS/Exchange.asmx
    at Request._callback (/Users/terenceyang/nylas-nodejs/lib/nylas-connection.js:193:23)
    at Request.self.callback (/Users/terenceyang/nylas-nodejs/node_modules/request/request.js:185:22)
    at Request.emit (events.js:314:20)
    at Request.<anonymous> (/Users/terenceyang/nylas-nodejs/node_modules/request/request.js:1161:10)
    at Request.emit (events.js:314:20)
    at IncomingMessage.<anonymous> (/Users/terenceyang/nylas-nodejs/node_modules/request/request.js:1083:12)
    at Object.onceWrapper (events.js:420:28)
    at IncomingMessage.emit (events.js:326:22)
    at endReadableNT (_stream_readable.js:1226:12)
    at processTicksAndRejections (internal/process/task_queues.js:80:21) {
  statusCode: 401
}
```

# License


src/nylas-connection.js
```diff
@@ -159,6 +159,9 @@ module.exports = class NylasConnection {
           }
           if (response.statusCode) {
             error.statusCode = response.statusCode;
+            if (!error.hasOwnProperty('message')) {
```
ozsivanov, line 162: Hi @Tranced, I think we may want to also check `typeof response.body === 'string'` to make sure we're returning a string. It's also worth considering adding an `else` clause in case `response.body` isn't a string, and just responding with a generic error message.

Tranced: Sounds good to me! I'm not sure what cases we'll run into if the error message is not a string. So, in that case, I'll just leave it with the current default behavior.

ozsivanov (approved):
LGTM 👍

PR #227, opened 2021-05-14
[58040] Added support for event metadata

mrashed-dev:
The Nylas Calendar API has a new beta feature of adding a new [metadata field to calendar events](https://www.nylas.com/blog/event-metadata). This PR makes the feature available via the Nylas Node.js SDK. 

# License

### DIFF
--- a/src/models/event.ts
+++ b/src/models/event.ts
@@ -27,6 +27,7 @@ export default class Event extends RestfulModel {
     rrule: string[];
     timezone: string;
   };
+  metadata?: object;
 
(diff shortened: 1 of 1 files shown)


mrashed-dev:
Testing was complete using an event (created with metadata) via both the API and via the SDK as well. Events can be created with metadata within the SDK and the SDK can also read metadata successfully.

philrenaud (approved):

AaronDDM (approved):

PR #233, opened 2021-05-27
[59706] Replace deprecated 'request' library

mrashed-dev:
# Description
The `requests` library has been deprecated for a while now. We've since moved on to `node-fetch` as it's supported elsewhere in the SDK. Test suite has been updated as well to make sure that we don't fail because of this change.

# License

### DIFF
--- a/__tests__/nylas-api-spec.js
+++ b/__tests__/nylas-api-spec.js
@@ -1,13 +1,13 @@
 jest.mock('node-fetch', () => {
-  const { Response } = jest.requireActual('node-fetch');
+  const { Request, Response } = jest.requireActual('node-fetch');
   const fetch = jest.fn();
+  fetch.Request = Request;
   fetch.Response = Response;
   return fetch;
 });
 
 // TODO since node 10 URL is global
 import { URL } from 'url';
-import request from 'request';
 import fetch, { Response } from 'node-fetch';
 import Nylas from '../src/nylas';
 import NylasConnection from '../src/nylas-connection';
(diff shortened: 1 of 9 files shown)


src/nylas-connection.ts
```diff
@@ -64,69 +61,61 @@ export default class NylasConnection {
...
-
-    if (options.formData) {
-      newOptions.formData = options.formData;
-    } else {
-      newOptions.body = options.body || {};
-    }
-
-    // For convenience, If `expanded` param is provided, convert to view:
-    // 'expanded' api option
+  requestOptions(options: RequestOptions): RequestOptions {
```
TrySound, line 64: Maybe combine with getRequest method I added specifically for fetch?

mrashed-dev: I actually addressed this but forgot to commit the code! They're very similar methods so now we have a `newRequest` method that returns the new `Request` object so that the delta class can do its own processing separate from what happens in the `request` method. Let me know what you think/comments.

__tests__/nylas-api-spec.js
```diff
@@ -247,29 +248,31 @@ describe('Nylas', () => {
...
-    test('should make a GET request to /a/<clientId> when options are not provided', () => {
-      request.Request = jest.fn(options => {
-        expect(options.url).toEqual('https://api.nylas.com/a/myId');
-        expect(options.auth.user).toEqual('mySecret');
-        expect(options.method).toEqual('GET');
-      });
-      Nylas.application();
-    });
-
-    test('should make a PUT request to /a/<clientId> when options are provided', () => {
```
AaronDDM, line 259: Strange functionality, but are these test cases not applicable anymore?

AaronDDM (approved):

mrashed-dev:
Closing due to rebase issue. Please see #234

PR #236, opened 2021-06-01
[60372] Return better error details from API

mrashed-dev:
# Description
Previously, we would only return generally the error message. Now we return a custom error object (that extends `Error`) that properly stores all of the error values we get back from the API and sets them in a way that is more native than just returning a string.

# Usage
The new `NylasApiError` class extends the `Error` class and both sets the build-in properties of Error and extends the error class to provide more values if needed. The following properties are made availalbe by the class:

- `name` (string): The name of the error, derived from the [API Errors page](https://developer.nylas.com/docs/developer-tools/api/errors/) (e.g. `Bad Request`)
- `message` (string): The error message returned from the API (e.g. `No recipients specified`)
- `statusCode` (number): The status code returned from the API call (e.g. `400`)
- `type` (string): The type of error returned from the API (e.g. `invalid_request_error`)
- `missingFields` (string[]): The array of fields that were missing in the call, if applicable (e.g. `["name", "email"]`)
- `serverError` (string): The error returned by the provider (e.g. `interaction_required`)
- `stack` (string): The stacktrace from JavaScript

# License

### DIFF
--- a/src/models/NylasApiError.ts
+++ b/src/models/NylasApiError.ts
@@ -1,0 +1,22 @@
+export default class NylasApiError extends Error {
+    statusCode: number;
+    missingFields?: string[];
+    serverError?: string;
+
+    constructor(status_code: number, type: string, message: string) {
+        super(message);
+        this.statusCode = status_code;
+        this.name = type;
+    }
+
+    toString = () => {
+        return {
+            type: this.name,
+            message: this.message,
+            statusCode: this.statusCode,
+            missingFields: this.missingFields,
+            serverError: this.serverError,
+            stack: this.stack
+        }
+    }
+}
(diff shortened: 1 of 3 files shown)


AaronDDM (approved):

src/models/NylasApiError.ts
```diff
@@ -0,0 +1,60 @@
...
+        return {
+            name: this.name,
+            message: this.message,
+            statusCode: this.statusCode,
+            type: this.type,
+            missingFields: this.missingFields,
+            serverError: this.serverError,
+            stack: this.stack
+        }
+    }
```
AaronDDM, line 59: `toString()` should return a string, which is typically for an `Error` the `message` that was passed in. 
So we technically do not need this method if it returns the value of `this.message`.

mrashed-dev: Good point. As discussed, for the sake of backwards compatibility and confusion I've removed this method. Thanks!

AaronDDM (approved):

PR #237, opened 2021-06-04
[59076] Add read only fields and enable v2.2 API support

mrashed-dev:
# Description
This PR was spawned by the need to add new read-only fields to the [Room Resource](https://developer.nylas.com/docs/api/#tag--Room-Resources--room-resources-object) object. We did not have support for read-only objects yet in this SDK, so this PR achieves the following:
- Enabling v2.2 API support
- Adding read-only field mechanism
- Add new read-only Room Resource fields

# License


src/models/webhook.ts
```diff
@@ -30,11 +30,6 @@ export default class Webhook extends ManagementModel {
   save(params: {} | SaveCallback = {}, callback?: SaveCallback) {
     return this._save(params, callback);
   }
-  toJSON() {
-    const json = super.toJSON();
-    delete json['object'];
-    return json;
-  }
```
AaronDDM, line 37: Is this a breaking change? Is there any concern?

mrashed-dev: Not really. This is just now redundant with the new changes I implemented in the PR. `object` is supposed to be readonly, and thus the old workaround was to delete the object before sending it. Now, because we have readonly attributes, when the parent `toJSON()` is called by `saveRequestBody` (which is called by the SDK when constructing a payload to send to the Nylas API) it will ignore the `object` key:
```
toJSON(enforceReadOnly?: boolean) { // saveReqeustBody sends in 'true'
  const json: any = {};
  const attributes = this.attributes();
  for (const attrName in attributes) {
    if (!enforceReadOnly || !attributes[attrName].readOnly) { // 'object'.readOnly === true
      const attr = attributes[attrName];
      json[attr.jsonKey] = attr.toJSON((this as any)[attrName]);
    }
  }
  return json;
}
```

AaronDDM (approved):

AaronDDM (approved):

PR #234, opened 2021-05-31
[59706] Replace deprecated 'request' library (Rebased)

mrashed-dev:
# Description
The `requests` library has been deprecated for a while now. We've since moved on to `node-fetch` as it's supported elsewhere in the SDK. Test suite has been updated as well to make sure that we don't fail because of this change.

Please see #233 for reviewer comments.

# License
I confirm that this contribution is made under the terms of the MIT license and that I have the authority necessary to make this contribution on behalf of its copyright owner.

PR #240, opened 2021-06-10
Move request to dev dependencies

TrySound:
Request package was left in dependencies here https://github.com/nylas/nylas-nodejs/pull/234

# License

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,8 +47,7 @@
     "backoff": "^2.5.0",
     "form-data": "^4.0.0",
     "JSONStream": "^1.3.5",
-    "node-fetch": "^2.6.1",
-    "request": "^2.88.0"
+    "node-fetch": "^2.6.1"
   },
   "devDependencies": {
     "@babel/cli": "^7.13.16",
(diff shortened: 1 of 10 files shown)


TrySound:
cc @mrashed-dev

package.json
```diff
@@ -68,6 +67,7 @@
     "jest": "^24.9.0",
     "prettier": "^1.16.4",
     "pretty-format": "^25.2.6",
+    "request": "^2.88.0",
```
mrashed-dev, line 70: Do we need to actually keep this? I'm fine with taking it out completely, not even having it as a dev dependency.  Same with line 60, `"@types/request": "^2.48.4"`, we can probably take both lines out!

mrashed-dev (commented):
Good catch! Left a comment let me know what you think @TrySound

package.json
TrySound: I still see it in tests

TrySound: Ah, I see. Unused in tests :)

TrySound: Removed

mrashed-dev (approved):
Great catch, thanks for the PR @TrySound!

PR #242, opened 2021-06-23
[62896] `undefined` returned when trying to download file

mrashed-dev:
# Description
When downloading a file, the body returns `undefined`. This was due to a previous change when we cut over from the old `requests` library and moved to the `node-fetch` library. Test suite has also been updated to ensure that we have coverage for this in the future.

# License

### DIFF
--- a/src/models/file.ts
+++ b/src/models/file.ts
@@ -75,16 +75,7 @@ export default class File extends RestfulModel {
         path: `/files/${this.id}/download`,
         downloadRequest: true,
       })
-      .then(response => {
-        let filename;
-        const file = { ...response.headers, body: response.body };
-        if ('content-disposition' in file) {
-          filename =
-            /filename=([^;]*)/.exec(file['content-disposition'])![1] ||
-            'filename';
-        } else {
-          filename = 'filename';
-        }
+      .then(file => {
         if (callback) {
           callback(null, file);
         }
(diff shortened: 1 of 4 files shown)


src/models/file.ts
```diff
@@ -75,16 +75,7 @@ export default class File extends RestfulModel {
         path: `/files/${this.id}/download`,
         downloadRequest: true,
       })
-      .then(response => {
-        let filename;
```
mrashed-dev, line 79: This is safe to delete because we never use the filename in this block nor do we return it.

src/nylas-connection.ts
```diff
@@ -197,7 +195,17 @@ export default class NylasConnection {
           });
         } else {
           if (options.downloadRequest) {
-            return resolve(response.body);
+            response.buffer().then(buffer => {
+              // Return an object with the headers and the body as a buffer
+              const fileDetails: { [key: string]: any } = {};
+              response.headers.forEach((v, k) => {
```
mrashed-dev, line 201: Not sure if this is necessary or not, but in the previous versions the SDK would return the headers in an object with the header, so I'm adding this to maintain backwards compatibility. Also `response.headers` is no longer a JS object but some sort of map so I can't just spread the headers into an object :/

philrenaud (approved):

AaronDDM (approved):

PR #247, opened 2021-06-25
Respect options.json a bit more to fix sending raw mime (fixes #246).

### LINKED ISSUE #246: v5.5 breaks "Send Raw Mime", RequestOption's json flag is not being honored
Version 5.5 is no longer respecting the `RequestOptions`'s `json` value. In `requestOptions` if `options.body` is set, it is automatically JSON stringified regardless of `options.json`: https://github.com/nylas/nylas-nodejs/blob/42a032db709dd172eb2b997fa2d7ed88b3323caa/src/nylas-connection.ts#L125

This appears to have been broke in this commit: https://github.com/nylas/nylas-nodejs/commit/b995b4f4489c09e05f07c810772b7d816399fe4e, when switching to node-fetch.

Because of this version 5.5 breaks "Send Raw Mime" as documented here: https://developer.nylas.com/docs/developer-tools/sdk/javascript-sdk/#send-raw-mime. As you can see, there is a `json` flag set to false when this.rawMime is set, and flipped to true otherwise:

https://github.com/nylas/nylas-nodejs/blob/42a032db709dd172eb2b997fa2d7ed88b3323caa/src/models/draft.ts#L78

When the request is made, the raw mime content is stringified to JSON and the resulting API call fails. To make matters worse, we've been passing a `Buffer` object which isn't JSON serializable. Even if I was passing a string for `options.body`, running it through `JSON.stringify` would still not work.

I tried to make a PR, by simply changing the logic to:

```typescript
if (options.formData) {
  ...
} else if (options.body && options.json) {
  options.body = JSON.stringify(options.body);
}
```

but this seemingly breaks some calendaring unit tests.

mmattax:
Version 5.5 has a breaking change where it calls `JSON.stringify` on `options.body` regardless of `options.json`. This breaks the [send raw mime](https://developer.nylas.com/docs/developer-tools/sdk/javascript-sdk/#send-raw-mime) functionality. I document this in a bit more detail here: https://github.com/nylas/nylas-nodejs/issues/246

Checking to see if options.json is explicitly set to `false` seems pretty safe here (rather than make the condition `options.body && options.json` as that breaks some calendar unit tests even though it's probably a bit more semantically correct). It looks like this library use to do something similar: https://github.com/nylas/nylas-nodejs/commit/b995b4f4489c09e05f07c810772b7d816399fe4e#diff-2de4f9bb82a400d2734c75d30d64883eec745110373fcde2c3e46fea7d3eb325L229

# License

### DIFF
--- a/src/nylas-connection.ts
+++ b/src/nylas-connection.ts
@@ -122,7 +122,7 @@ export default class NylasConnection {
         }
       }
       options.body = fd;
-    } else if (options.body) {
+    } else if (options.body && options.json !== false) {
       options.body = JSON.stringify(options.body);
     }
 


mrashed-dev:
Great PR! Thanks for this fix, I just ran into this issue myself as I was fixing some of our test cases that were passing prematurely #249. Looks good to me, merging in - thanks again!

mrashed-dev (approved):
Tested and works, LGTM!

PR #253, opened 2021-07-08
[63117] Neural API Support

mrashed-dev:
# Description
This PR brings support for the Nylas [Neural API](https://developer.nylas.com/docs/intelligence/). The Neural API exposes endpoints that perform intelligent analysis on items in your Nylas account. This change will enable users to access the features of the Nylas Neural API and provides some utility functions to help improve the user's workflow.

# Usage
To use [Sentiment Analysis](https://developer.nylas.com/docs/intelligence/sentiment-analysis/):
```
// To perform sentiment analysis on a message, pass in the message ID:
const messageAnalysis = await nylas.neural.sentimentAnalysisMessage(MESSAGE_ID);

// To perform sentiment analysis on just text, pass in a string:
const textAnalysis = await nylas.neural.sentimentAnalysisText("Hi, thank you so much for reaching out! We can catch up tomorrow.");
```

To use [Signature Extraction](https://developer.nylas.com/docs/intelligence/signature-extraction/):
```
const signature = await nylas.neural.extractSignature(MESSAGE_ID);

// The method also accepts two optional parameters
// parseContact, a boolean for whether Nylas should parse the contact from the signature (API defaults to true)
// options, an object of options that can be enabled for the Neural endpoint, of type NeuralMessageOptions:
const options = {
    ignore_links: false,
    ignore_images: false,
    ignore_tables: false,
    remove_conclusion_phrases: false,
    images_as_markdowns: false
}

const signature = await nylas.neural.extractSignature(MESSAGE_ID, true, options);
```
and to parse the contact and convert it to the standard Nylas contact object:
```
const contact = signature.contacts.toContactObject();
```

To use [Clean Conversations](https://developer.nylas.com/docs/intelligence/clean-conversations/):
```
const convo = await nylas.neural.cleanConversation(MESSAGE_ID);

// You can also pass in an object of options that can be enabled for the Neural endpoint, of type NeuralMessageOptions
const convo = await nylas.neural.cleanConversation(MESSAGE_ID, options);
```
and to extract images from the result:
```
await convo.extractImages();
```

To use [Optical Character Recognition](https://developer.nylas.com/docs/intelligence/optical-charecter-recognition/):
```
const ocr = await nylas.neural.ocrRequest(FILE_ID);

// This endpoint also supports a second, optional parameter for an array specifying the pages that the user wants analyzed:
const ocr = await nylas.neural.ocrRequest(FILE_ID, [2,3]);
```

To use [Categorizer](https://developer.nylas.com/docs/intelligence/categorizer/)
```
let cat = await nylas.neural.categorize(MESSAGE_ID);

// You can also send a request to recategorize the message:
cat = await cat.reCategorize("conversation");
```

# License


philrenaud (approved):

src/models/neural-clean-conversation.ts
```diff
@@ -0,0 +1,36 @@
...
+import File from './file';
+
+export default class NeuralCleanConversation extends Message {
+  conversation?: string;
+  modelVersion?: string;
+
+  extractImages(): Promise<File[]> {
+    const f: File[] = [];
+    if (this.conversation) {
+      const regex = /[(']cid:(.)*[)']/g;
```
sdwvit, line 12: would be nice to move this on the top of the file

tjperry07 (approved):
Love the examples in the comments! Makes it so much easier for me to write documentation.

src/models/neural-clean-conversation.ts
```diff
@@ -0,0 +1,36 @@
...
+      const fileIds = this.conversation.match(regex);
+      if (fileIds) {
+        fileIds.forEach(async id => {
+          const parsedId = id.substring(5, id.length - 1);
```
sdwvit, line 16: would be nice to have a comment about how id looks like

src/models/neural-sentiment-analysis.ts
```diff
@@ -0,0 +1,36 @@
...
+import RestfulModel from './restful-model';
+
+export default class NeuralSentimentAnalysis extends RestfulModel {
+  account_id?: string;
+  sentiment?: string;
+  sentimentScore?: number;
+  processedLength?: number;
+  text?: string;
+
+  fromJSON(json: object): this {
```
sdwvit, line 11: how about 
```suggestion
  fromJSON(json: unknown): this {
```

sdwvit: or typeof this

src/models/neural.ts
```diff
@@ -0,0 +1,137 @@
...
+        ...body,
+        ...options,
+      };
+    }
+    if (parseContact) {
+      body['parse_contact'] = parseContact;
+    }
+
+    return this._request(path, body).then(json => {
+      if (Array.isArray(json)) {
```
sdwvit, line 49: in which cases there may be more elements?

sdwvit (approved):

arjunkalburgi (approved):

src/models/neural-sentiment-analysis.ts
mrashed-dev: @sdwvit just curious, why going from typing object to unknown?

src/models/neural.ts
mrashed-dev: According to the docs, never, but sometimes the API response that gets returned is in the form of an Array of just 1  element. Here's an example: https://developer.nylas.com/docs/intelligence/signature-extraction/#parse-signature-response

src/models/neural-clean-conversation.ts
mrashed-dev: Good catch!

src/models/neural-clean-conversation.ts
mrashed-dev: Good idea.... I should probably have better comments in general

src/models/neural-sentiment-analysis.ts
sdwvit: object is as broad as any in this case. Unknown just means you don't know the type. 
Not the best idea to have either of those. 

How about: 
```suggestion
  fromJSON(json: Partial<typeof this>): this {
```

sdwvit: or anything like that, that imrpoves type

mrashed-dev: I love the suggestion! Actually it turns out I don't even need this piece of code, it's redundant all it does is call the super method and the class `extends RestfulModel` so there's no need for it.

mrashed-dev:
Thanks for everyone's reviews! I will get a new release with this functionality out sometime tomorrow! 🎉

PR #263, opened 2021-08-09
[64881] Test documenting with documentation.js

mrashed-dev:
# Description
`documentation.js` is now a dev dependency to generate documentation based on comments. 

# Usage
To generate documents, just run `npm run docs` in the root directory.

# License


mrashed-dev:
Only opened for a test, no longer needed.

PR #264, opened 2021-08-13
[65378] Dynamic conferencing link creation support

mrashed-dev:
# Description
The Event API, in addition to support for manually-added conference details, supports auto generating conference details. If you have an integration authorized and enabled for your Nylas account you can provide an `autocreate` object instead of a `details` object to have the Nylas API autofill the conferencing details.

# Usage
To have Nylas autocreate the conference field for you, pass the `autocreate` object to the new event:
```javascript
const zoomEvent = nylas.events.build({
    ...
    conferencing: {
        provider: "Zoom Meeting",
        autocreate: {
            settings: {},
        }
    },
});
```
A few notes and things to keep in mind:
* Only **one** of `details` or `autocreate` can be present, and we have implemented client-side checking to enforce this rule
* Autocreating conferencing data is an asynchronous operation on the server-side. The Event object returned will not have a conferencing field, but it should be available in a future get call once the conference is created on the backend. The initial Event object returned will have a `jobStatusId` value which can be used to check on the status of conference creation.
* The `settings` object within the `autocreate` object maps to the settings the Nylas API will send to the conference provider for the conference creation. For example with Zoom the settings object maps to Zoom's [Create a Meeting object](https://marketplace.zoom.us/docs/api-reference/zoom-api/meetings/meetingcreate).

# License

### DIFF
--- a/src/models/event-conferencing.ts
+++ b/src/models/event-conferencing.ts
@@ -41,11 +41,15 @@ EventConferencingDetails.attributes = {
 export class EventConferencing extends RestfulModel {
   details?: EventConferencingDetails;
   provider?: string;
+  autocreate?: {
+    settings?: object;
+  }
 
   toJSON(): any {
     return {
       details: this.details,
       provider: this.provider,
+      autocreate: this.autocreate,
     };
   }
 }
@@ -57,5 +61,8 @@ EventConferencing.attributes = {
   }),
   provider: Attributes.String({
     modelKey: 'provider',
+  }),
+  autocreate: Attributes.Object({
+    modelKey: 'autocreate',
   }),
 };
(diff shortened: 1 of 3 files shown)


src/models/event-conferencing.ts
```diff
@@ -41,11 +41,15 @@ EventConferencingDetails.attributes = {
 export class EventConferencing extends RestfulModel {
   details?: EventConferencingDetails;
   provider?: string;
+  autocreate?: {
+    settings?: object;
```
ozsivanov, line 45: Is `settings` optional?

mrashed-dev: @ozsivanov Yes it is!

ozsivanov (approved):
LGTM 👍

philrenaud (approved):

PR #267, opened 2021-08-20
[66695] Refactor Restfulmodel

mrashed-dev:
# Description
As the SDK coverage grows, we need to decouple the standard functionality of a model from the specific functionality/structure of a "classic" Nylas API model. A "classic" Nylas API model is a model used for endpoints like `Event`, `Calendar`, `Message`, etc. where we expect the object to have an `id`, `account_id`, and `object` field. There are a lot of "dumb" models like `Contact.Emails` that would benefit from being cast as a basic model versus the current RestfulModel, as well as having the ability in the future for creating different types of models.

# License

### DIFF
--- a/src/models/attributes.ts
+++ b/src/models/attributes.ts
@@ -1,3 +1,4 @@
+import Model from './model';
 import RestfulModel from './restful-model';
 
 // The Attribute class represents a single model attribute, like 'namespace_id'
@@ -33,7 +34,7 @@ export class Attribute {
 }
 
 class AttributeObject extends Attribute {
-  itemClass?: typeof RestfulModel;
+  itemClass?: typeof Model | typeof RestfulModel;
 
   constructor({
     modelKey,
@@ -43,7 +44,7 @@ class AttributeObject extends Attribute {
   }: {
     modelKey: string;
     jsonKey?: string;
-    itemClass?: typeof RestfulModel;
+    itemClass?: typeof Model | typeof RestfulModel;
     readOnly?: boolean;
   }) {
     super({ modelKey, jsonKey, readOnly });
@@ -64,7 +65,11 @@ class AttributeObject extends Attribute {
     if (!val || !this.itemClass) {
       return val;
     }
-    return new this.itemClass(_parent.connection, val);
+
+    if (this.itemClass.prototype instanceof RestfulModel) {
+      return new this.itemClass(_parent.connection, val);
+    }
+    return new this.itemClass(val);
   }
 }
 
@@ -155,7 +160,7 @@ class AttributeDateTime extends Attribute {
 }
 
 class AttributeCollection extends Attribute {
-  itemClass: typeof RestfulModel;
+  itemClass: typeof Model | typeof RestfulModel;
 
   constructor({
     modelKey,
@@ -165,7 +170,7 @@ class AttributeCollection extends Attribute {
   }: {
     modelKey: string;
     jsonKey?: string;
-    itemClass: typeof RestfulModel;
+    itemClass: typeof Model | typeof RestfulModel;
     readOnly?: boolean;
   }) {
     super({ modelKey, jsonKey, readOnly });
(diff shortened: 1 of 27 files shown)


src/models/attributes.ts
```diff
@@ -193,7 +198,12 @@ class AttributeCollection extends Attribute {
     }
     const objs = [];
     for (const objJSON of json) {
-      const obj = new this.itemClass(_parent.connection, objJSON);
+      let obj;
+      if (this.itemClass.prototype instanceof RestfulModel) {
```
ozsivanov, line 202: I like this pattern, but I think it can be improved a bit by having `RestfulModel`'s constructor accept the connection object as the second prop instead of the first. That way, you don't need these `if` checks and can just try to instantiate it like this.
```
obj = new this.itemClass(objJSON, _parent.connection);
```
If `itemClass` is of type `Model`, it would just ignore the second param in its constructor. The downside of doing this is that we'd have to implement some conditional logic if `Model` ever requires a second param in the future.

src/models/model.ts
```diff
@@ -0,0 +1,44 @@
...
+export default class Model {
+  static attributes: { [key: string]: Attribute };
+
+  constructor(json?: any) {
+    if (json) {
+      this.fromJSON(json);
+    }
+  }
+
+  attributes(): { [key: string]: Attribute } {
```
ozsivanov, line 14: Why is this needed?

src/models/model.ts
```diff
@@ -0,0 +1,44 @@
...
+    return (this.constructor as any).attributes;
+  }
+
+  fromJSON(json?: any): Model {
+    const attributes = this.attributes();
+    for (const attrName in attributes) {
+      const attr = attributes[attrName];
+      if (json[attr.jsonKey] !== undefined) {
+        (this as any)[attrName] = attr.fromJSON(json[attr.jsonKey], this);
```
ozsivanov, line 23: I think you should be able to have `(this as Model)` here

src/models/attributes.ts
mrashed-dev: That's actually a really clean idea! My only concern is that the objJSON parameter (json) is optional for both `RestfulModel` and `Model` so I can't make it the first parameter on `RestfulModel` unless I make `connection: NylasConnection` optional (which is not ideal since we need this parameter).

src/models/model-collection.ts
```diff
@@ -0,0 +1,251 @@
...
+      })
+      .catch(err => {
+        if (callback) {
+          callback(err);
+        }
+        return Promise.reject(err);
+      });
+  }
+
+  _range({
```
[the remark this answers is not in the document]
mrashed-dev: Sounds good, I applied `private` and `protected` method modifiers for all methods that followed `_[function name]` notation

src/models/model.ts
```diff
@@ -0,0 +1,44 @@
+import { Attribute } from './attributes';
+
+export type SaveCallback = (error: Error | null, result?: any) => void;
+
+export default class Model {
+  static attributes: { [key: string]: Attribute };
```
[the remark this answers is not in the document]
mrashed-dev: I was toying with the idea of abstracting classes and even variables like this one but I ran into some issues:

- when making `Model` abstract and you try to create an instance of class `typeof Model` typescript throws an error "Cannot create an instance of an abstract class" even though you are trying to instantiate a class `typeof Model`, not directly of `Model`
- Making `attributes` static means making `Model` static  and every class extending `Model` must implement their own `attributes`, but the implementation would all be the same so it would be a duplicate implementation across all classes that extend `Model`

So given the added complexity of making it abstract I decided against it

src/models/model.ts
mrashed-dev: I don't think it's _needed_ I think it's just there to make the code look cleaner by inserting `this.attributes()` instead of `(this.constructor as any).attributes`.

src/models/model.ts
mrashed-dev: I tried that but it throws an error complaining about using a `string` as a index for type `Model`

> TS7053: Element implicitly has an 'any' type because expression of type 'string' can't be used to index type 'Model'. No index signature with a parameter of type 'string' was found on type 'Model'.

One way around this would be to have an interface like `export interface ModelProperties { [key: string]: any }` and use it like: `(this as ModelProperties)[attrName] = attr.fromJSON(json[attr.jsonKey], this);` but I'm not sure if that's any better.

philrenaud (approved):

src/models/attributes.ts
ozsivanov: Oh I didn't realize they were optional, let's just leave it as is then

ozsivanov (approved):

PR #274, opened 2021-09-14
[63533] Improve Node SDK typing

mrashed-dev:
# Description
This PR is to address the Node SDK typing issues our customers have been experiencing. The biggest change we've made here is that all of the SDK methods that are meant to be used by the user for Nylas operations all have a non-any return type when possible.

In addition to this, the PR improves on the previous work of the refactor to ensure that we are utilizing the use of models, interfaces, and proper deserialization. With this, our support for Calendar availability has improved with the introduction of classes and types for API objects like free-busy and availability. Our support for Native Authentication and Virtual Calendars have also seen improvements as we have models to represent these objects and introduced enums to provide the user an easier time with providing things like providers and scopes without needing to refer to the API docs to see what scopes/providers are supported and to reduce user typing errors.

# License

### DIFF
--- a/__tests__/connect-spec.js
+++ b/__tests__/connect-spec.js
@@ -1,16 +1,23 @@
 import Nylas from '../src/nylas';
+import { NativeAuthenticationProvider, Scope } from '../src/models/connect';
 
 describe('Connect', () => {
   const name = 'Connect Test';
   const password = 'connecting123';
-  const scopes = 'email.modify,email.send,calendar,contacts';
+  const scopeString = 'email.modify,email.send,calendar,contacts';
+  const scopes = [
+    Scope.EmailModify,
+    Scope.EmailSend,
+    Scope.Calendar,
+    Scope.Contacts,
+  ];
   const exchangeEmail = 'connect_test@outlook.com';
   const CLIENT_ID = 'abc';
   const CLIENT_SECRET = 'xyz';
   const authorizeOptions = (email, provider, settings) => {
     return {
       name: name,
-      email_address: email,
+      emailAddress: email,
       provider: provider,
       settings: settings,
       scopes: scopes,
@@ -43,7 +50,11 @@ describe('Connect', () => {
       const settings = { username: exchangeEmail, password: password };
       expect(() =>
         Nylas.connect.authorize(
-          authorizeOptions(exchangeEmail, 'exchange', settings)
+          authorizeOptions(
+            exchangeEmail,
+            NativeAuthenticationProvider.Exchange,
+            settings
+          )
         )
       ).toThrow();
       expect(() => Nylas.connect.token(authorizeJSON.code)).toThrow();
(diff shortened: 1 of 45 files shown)


src/models/event.ts
```diff
@@ -136,15 +136,20 @@ export default class Event extends RestfulModel {
     }
   }
 
-  deleteRequestQueryString(params: { [key: string]: any } = {}) {
+  deleteRequestQueryString(
+    params: { [key: string]: any } = {}
```
AaronDDM, line 140: @mrashed-dev for such instances like `{ [key: string]: any }` - I think it might be better to opt to use the [Record<keyType, valueType>](https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeystype) built-in type.

For example `params: Recod<string,any>`.

AaronDDM: Separately, is there no way to avoid the use of `any`? It seems to defeat the purpose of typing :'(

If we do have to use `any`, could maybe [unknown](https://mariusschulz.com/blog/the-unknown-type-in-typescript) be of better use? (see article for more info)

src/models/model.ts
```diff
@@ -36,7 +36,7 @@ export default class Model {
     }
   }
 
-  fromJSON(json?: any): this {
+  fromJSON(json: { [key: string]: any }): any {
```
AaronDDM, line 39: Previously we returned `this`, but changing it to `any` would break typing for chaining, would it not?

e.g. 
```js
model.fromJSON({ ... }).toJSON()
                         ^ typing would be broken and unable determine if `toJSON()` is valid
```

philrenaud (approved):

__tests__/connect-spec.js
```diff
@@ -78,7 +95,7 @@ describe('Connect', () => {
               email_address: exchangeEmail,
```
AaronDDM, line 95: Some places we have `emailAddress` and some places `email_address`, is it possible to have this consistent throughout?

mrashed-dev: This assertion (`toHaveBeenCalledWith`) checks the outgoing POST call. In the TS code we use `emailAddress` (camelCase) as the property name but when it gets serialized to JSON it becomes email_address (snake_case) to conform with what the API expects.

src/models/event.ts
mrashed-dev: @AaronDDM thanks for bringing unknown to my attention! Let me play with it and see how it works with the different methods we have.

src/models/message.ts
```diff
@@ -75,11 +75,11 @@ export default class Message extends RestfulModel implements MessageProperties {
     return Object.values(participants);
   }
 
-  fileIds() {
+  fileIds(): (string | undefined)[] {
     return this.files ? this.files.map(file => file.id) : [];
   }
```
AaronDDM, line 80: @mrashed-dev should this really be an array of `string | undefined` or can we filter out `undefined` ids? Not sure if there is a requirement to get the same # of array elements even if we don't have valid IDs.
```suggestion
  fileIds(): (string | undefined)[] {
    return this.files ? this.files.filter(file => file.id).map(file => file.id) : [];
  }
```

src/models/event.ts
mrashed-dev: @AaronDDM I've updated the PR to include both changes!

src/models/model.ts
mrashed-dev: @AaronDDM good catch! Not sure why I reverted it. In any case, it's back in now.

src/models/message.ts
mrashed-dev: @AaronDDM I think that it will always return `string | undefined` because `this.files` is not guaranteed to be populated, which I think means it yields an undefined array type.

AaronDDM (approved):

PR #275, opened 2021-09-15
[66508] Add support consecutive availability

mrashed-dev:
# Description
This PR enables support for the `/calendars/availability/consecutive` endpoint. 

# Usage
To check for consecutive availability:
```javascript
const params = {
  startTime: 1590454800,
  endTime: 1590780800,
  interval: 5,
  duration: 30,
  buffer: 30,
  emails: [['jane@email.com'], ['swag@nylas.com']],
  open_hours: [
    {
      emails: ['swag@nylas.com', 'jane@email.com'],
      days: [0],
      timezone: 'America/Chicago',
      start: '10:00',
      end: '14:00',
      object_type: 'open_hours',
    },
  ],
};

const consecutiveAvailability = await nylas.calendars.consecutiveAvailability(params);
```

# License

### DIFF
--- a/__tests__/calendar-restful-model-collection-spec.js
+++ b/__tests__/calendar-restful-model-collection-spec.js
@@ -202,7 +202,7 @@ describe('CalendarRestfulModelCollection', () => {
           free_busy: [],
           open_hours: [
             {
-              emails: ['swag@nylas.com'],
+              emails: [{ name: 'Nylas', email: 'swag@nylas.com' }],
               days: ['0'],
               timezone: 'America/Chicago',
               start: '10:00',
--- a/__tests__/calendar-restful-model-collection-spec.js
+++ b/__tests__/calendar-restful-model-collection-spec.js
@@ -173,16 +173,20 @@ describe('CalendarRestfulModelCollection', () => {
       interval: 5,
       duration: 30,
       emails: [['jane@email.com'], ['swag@nylas.com']],
-      free_busy: [{
-        email: 'jane@email.com',
-        object: 'free_busy',
-        time_slots: [{
-          object: "time_slots",
-          status: "busy",
-          start_time: 1590454800,
-          end_time: 1590780800,
-        }]
-      }],
+      free_busy: [
+        {
+          email: 'jane@email.com',
+          object: 'free_busy',
+          time_slots: [
+            {
+              object: 'time_slots',
+              status: 'busy',
+              start_time: 1590454800,
+              end_time: 1590780800,
+            },
+          ],
+        },
+      ],
       open_hours: [
         {
           emails: ['jane@email.com', 'swag@nylas.com'],
@@ -209,16 +213,20 @@ describe('CalendarRestfulModelCollection', () => {
           interval_minutes: 5,
           duration_minutes: 30,
           emails: [['jane@email.com'], ['swag@nylas.com']],
-          free_busy: [{
-            email: 'jane@email.com',
-            object: 'free_busy',
-            time_slots: [{
-              object: "time_slots",
-              status: "busy",
-              start_time: 1590454800,
-              end_time: 1590780800,
-            }]
-          }],
+          free_busy: [
+            {
+              email: 'jane@email.com',
+              object: 'free_busy',
+              time_slots: [
+                {
+                  object: 'time_slots',
+                  status: 'busy',
+                  start_time: 1590454800,
+                  end_time: 1590780800,
+                },
+              ],
+            },
+          ],
           open_hours: [
             {
               emails: ['jane@email.com', 'swag@nylas.com'],
@@ -246,16 +254,20 @@ describe('CalendarRestfulModelCollection', () => {
       interval: 5,
       duration: 30,
       emails: [['jane@email.com']],
-      free_busy: [{
-        email: 'jane@email.com',
-        object: 'free_busy',
-        time_slots: [{
-          object: "time_slots",
-          status: "busy",
-          start_time: 1590454800,
-          end_time: 1590780800,
-        }]
-      }],
+      free_busy: [
+        {
+          email: 'jane@email.com',
+          object: 'free_busy',
+          time_slots: [
+            {
+              object: 'time_slots',
+              status: 'busy',
+              start_time: 1590454800,
+              end_time: 1590780800,
+            },
+          ],
+        },
+      ],
       open_hours: [
         {
           emails: ['jane@email.com', 'swag@nylas.com'],
@@ -268,7 +280,9 @@ describe('CalendarRestfulModelCollection', () => {
       ],
     };
 
-    expect(() => testContext.connection.calendars.consecutiveAvailability(params)).toThrow();
+    expect(() =>
+      testContext.connection.calendars.consecutiveAvailability(params)
+    ).toThrow();
     done();
   });
 
--- a/__tests__/calendar-restful-model-collection-spec.js
+++ b/__tests__/calendar-restful-model-collection-spec.js
@@ -175,7 +175,7 @@ describe('CalendarRestfulModelCollection', () => {
       emails: [{ name: 'Jane', email: 'jane@email.com' }],
       open_hours: [
         {
-          emails: ['swag@nylas.com'],
+          emails: [{ name: 'Nylas', email: 'swag@nylas.com' }],
           days: ['0'],
           timezone: 'America/Chicago',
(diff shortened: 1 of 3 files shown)


philrenaud (approved):
cc @tbadun  / @jseller  -- could you give the logic a review and verify that this looks as expected?

__tests__/calendar-restful-model-collection-spec.js
```diff
@@ -166,6 +166,60 @@ describe('CalendarRestfulModelCollection', () => {
     });
   });
 
+  test('[CONSECUTIVE AVAILABILITY] should fetch results with params', done => {
+    const params = {
+      startTime: '1590454800',
+      endTime: '1590780800',
+      interval: 5,
+      duration: 30,
+      emails: [{ name: 'Jane', email: 'jane@email.com' }],
```
tbadun, line 175: currently the API just accepts lists of lists of emails, e.g.,
```
[["jane@email.com" ],["bill@email.com","ted@email.com"]]
```
looks for consecutive slots where jane is free and biill and ted are both free

__tests__/calendar-restful-model-collection-spec.js
```diff
@@ -166,6 +166,60 @@ describe('CalendarRestfulModelCollection', () => {
...
+      open_hours: [
+        {
+          emails: [{ name: 'Nylas', email: 'swag@nylas.com' }],
```
tbadun, line 178: similar here - plus you'll get an error of the open_hours contains an email that does not exist in the email or free_busy lists

tbadun: unless there is some other parsing step idk

__tests__/calendar-restful-model-collection-spec.js
mrashed-dev: @tbadun Ah gotcha, I think the documentation is wrong since that's where I got this from: https://developer.nylas.com/docs/api/#post/calendars/availability/consecutive

So just curious, will the first array (jane) always only have one email where as the second array has all the other participants we want to check against?

__tests__/calendar-restful-model-collection-spec.js
mrashed-dev: @tbadun thank you for letting me know about that error, I'll look into maybe having a check to ensure that open_hours emails all match the emails as found in the emails element in the main object. Is the email in the `open_hours` also a 2D array for consecutive?

__tests__/calendar-restful-model-collection-spec.js
tbadun: yeah, the docs are a bit off unfortunately, I flagged it with doc team a while ago but should follow up.

within the array of arrays, you can have as many arrays as you want, each sub-array will contain as many emails as are required to check for a given meeting. 

```
"emails": [["jane@email.com" ],["bill@email.com"],["ted@email.com"]]
```
will look for 3 consecutive timeslots: one with jane, one with bill one with ted (arbitrary order)

```
"emails": [["jane@email.com" ,"bill@email.com","ted@email.com"]]
```
will look for 1 timeslot where jane, bill, and ted are all available

basically, any sub-array will check for the availability of just those individuals.

__tests__/calendar-restful-model-collection-spec.js
```diff
@@ -166,6 +166,60 @@ describe('CalendarRestfulModelCollection', () => {
...
+          days: ['0'],
```
tbadun, line 179: this is also wrong in the docs, this should be a list of integer(s)

__tests__/calendar-restful-model-collection-spec.js
```diff
@@ -166,6 +166,60 @@ describe('CalendarRestfulModelCollection', () => {
     });
   });
 
+  test('[CONSECUTIVE AVAILABILITY] should fetch results with params', done => {
+    const params = {
+      startTime: '1590454800',
+      endTime: '1590780800',
```
tbadun, line 172: also both integers

src/models/calendar-restful-model-collection.ts
```diff
@@ -105,4 +105,68 @@ export default class CalendarRestfulModelCollection extends RestfulModelCollecti
...
+        path: `/calendars/availability/consecutive`,
+        body: {
+          emails: options.emails,
+          duration_minutes: options.duration,
+          interval_minutes: options.interval,
+          start_time: options.startTime || options.start_time,
+          end_time: options.endTime || options.end_time,
+          free_busy: options.free_busy || [],
+          open_hours: options.open_hours,
+        },
```
tbadun, line 157: You can also add `buffer` as a potential input to the call although it's not in the docs yet. It takes a non-negative integer or float representing the buffer minutes to add to busy time. 

If, for example, you send `buffer: 30`, there will be a 30 minute buffer before and after each individual's existing meetings/busy time that will be counted as busy.

src/models/calendar-restful-model-collection.ts
```diff
@@ -105,4 +105,68 @@ export default class CalendarRestfulModelCollection extends RestfulModelCollecti
...
+        }>;
+        days: string[];
+        timezone: string;
+        start: string;
+        end: string;
+        object_type: string;
+      }>;
+    },
+    callback?: (error: Error | null, data?: { [key: string]: any }) => void
+  ) {
```
tbadun, line 144: in case i missed any in the notes above:
```suggestion
  consecutiveAvailability(
    options: {
      emails: Array<string[]>;
      duration: number;
      interval: number;
      start_time?: number;
      startTime?: number;
      end_time?: number;
      endTime?: number;
      free_busy?: Array<{
        email: string;
        object: string;
        time_slots: Array<{
          object: string;
          status: string;
          start_time: number;
          end_time: number;
        }>;
      }>;
      open_hours: Array<{
        emails: Array<string[]>;
        days: number[];
        timezone: string;
        start: number;
        end: number;
        object_type: string;
      }>;
    },
    callback?: (error: Error | null, data?: { [key: string]: any }) => void
  ) {
```

__tests__/calendar-restful-model-collection-spec.js
tbadun: in open_hours emails will just be a list/1D array

__tests__/calendar-restful-model-collection-spec.js
mrashed-dev: Ahhhh perfect explanation, thanks!

__tests__/calendar-restful-model-collection-spec.js
mrashed-dev: Fantastic, thanks!

src/models/calendar-restful-model-collection.ts
mrashed-dev: Added in, thank you!

src/models/calendar-restful-model-collection.ts
mrashed-dev: Yup, got it in, thank you!

src/models/calendar-restful-model-collection.ts
```diff
@@ -105,4 +105,80 @@ export default class CalendarRestfulModelCollection extends RestfulModelCollecti
...
+    }
+
+    return this.connection
+      .request({
+        method: 'POST',
+        path: `/calendars/availability/consecutive`,
+        body: {
+          emails: options.emails,
+          duration_minutes: options.duration,
+          interval_minutes: options.interval,
```
tbadun, line 163: does it make sense to add other naming options?, e.g., 
```suggestion
          duration_minutes: options.duration | options.duration_minutes | options.durationMinutes,
          interval_minutes: options.interval | options.interval_minutes | options.intervalMinutes,
```

tbadun (approved):
lgtm - love the new test for open_hours, maybe an expansion of duration/interval input names if it makes sense but not necessary

src/models/calendar-restful-model-collection.ts
mrashed-dev: Technically, yes, but this is the "old" way of handling camelCase vs. snake_case. Our new refactor (latest PR is #274) will be opting for camelCase always while searlizing/deserializing appropriately. For the sake of consistency of the current way the SDK does things I've opted to keep it the same for now.

tbadun: perfect - looks like you're gtg then 😄

PR #279, opened 2021-09-17
Updating exchange token response to match API

wallacecs007:
https://app.shortcut.com/nylas/story/69505/nylas-exchangecodefortoken-response-body-doesn-t-match-the-api-response

A customer noticed that we were returning only the access token in the exchange token response. This doesn't match the API's response so I updated the response body to include the full response body from the call we make to exchange the token. This will return identifying information about the account as well.

https://developer.nylas.com/docs/api/#post/oauth/token

{
access_token: string
account_id: string
email_address: string
provider: enum
token_type: string
}

# License


mrashed-dev:
Looks good to me! Will be included in the refactor!

PR #292, opened 2021-12-03
[74829] Fix bug where updating an Event results in an API error

mrashed-dev:
# Description
A recent API update caused updating an event using the SDK to return an error because the SDK sends a Participant's status as well as an object key within the Event.when object. 

# License

### DIFF
--- a/src/models/attributes.ts
+++ b/src/models/attributes.ts
@@ -30,6 +30,12 @@ export class Attribute {
   fromJSON(val: any, _parent: any) {
     return val || null;
   }
+  saveRequestBody(val: any) {
+    if(this.readOnly) {
+      return undefined;
+    }
+    return this.toJSON(val);
+  }
 }
 
 class AttributeObject extends Attribute {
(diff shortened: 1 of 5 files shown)


src/models/attributes.ts
```diff
@@ -30,6 +30,12 @@ export class Attribute {
   fromJSON(val: any, _parent: any) {
     return val || null;
   }
+  saveRequestBody(val: any) {
+    if (this.readOnly) {
+      return undefined;
```
ozsivanov, line 35: nit
```suggestion
      return;
```

src/models/attributes.ts
```diff
@@ -66,6 +74,13 @@ class AttributeObject extends Attribute {
     }
     return new this.itemClass(_parent.connection, val);
   }
+
+  saveRequestBody(val: any) {
+    if (this.readOnly) {
+      return undefined;
```
ozsivanov, line 80:
```suggestion
      return;
```

src/models/attributes.ts
```diff
@@ -198,6 +215,13 @@ class AttributeCollection extends Attribute {
     }
     return objs;
   }
+
+  saveRequestBody(val: any) {
+    if (this.readOnly) {
+      return undefined;
```
ozsivanov, line 221:
```suggestion
      return;
```

ozsivanov (approved):

PR #296, opened 2021-12-15
[76422] Add missing `provider` option for `urlForAuthentication`

mrashed-dev:
# Description
The `oauth/authorize` endpoint supports a `provider` option for forcing authentication against a different provider, this PR adds it in.

# License

### DIFF
--- a/__tests__/nylas-api-spec.js
+++ b/__tests__/nylas-api-spec.js
@@ -211,6 +211,17 @@ describe('Nylas', () => {
       );
     });
 
+    test('should use a provider when provided in the options', () => {
+      const options = {
+        loginHint: 'ben@nylas.com',
+        redirectURI: 'https://localhost/callback',
+        provider: 'icloud',
+      };
+      expect(Nylas.urlForAuthentication(options)).toEqual(
+        'https://api.nylas.com/oauth/authorize?client_id=newId&response_type=code&login_hint=ben@nylas.com&redirect_uri=https://localhost/callback&provider=icloud'
+      );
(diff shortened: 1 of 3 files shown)


__tests__/nylas-api-spec.js
```diff
@@ -211,6 +211,17 @@ describe('Nylas', () => {
       );
     });
 
+    test('should use a provider when provided in the options', () => {
+      const options = {
+        loginHint: 'ben@nylas.com',
+        redirectURI: 'https://localhost/callback',
+        provider: 'icloud',
+      };
+      expect(Nylas.urlForAuthentication(options)).toEqual(
```
ozsivanov, line 220: nit - For tests like this, I'd recommend using string composition in the expected value to make sure your tests don't break if you need to change your options in the future.

eg.
```
 const loginHint = 'ben@nylas.com',
     redirectURI = 'https://localhost/callback',
     provider = 'icloud';

 const options = {
      loginHint,
      redirectURI,
      provider,
 };
 
 expect(Nylas.urlForAuthentication(options))
     .toEqual(`https://api.nylas.com/oauth/authorize?client_id=newId&response_type=code&login_hint=${loginHint}&redirect_uri=${redirectURI}&provider=${provider}`
```

ozsivanov (approved):

__tests__/nylas-api-spec.js
mrashed-dev: Thanks! Updated the test cases to follow this format!

PR #306, opened 2022-01-17
[78781] Support new `calendars` field in free-busy/availability queries

mrashed-dev:
# Description
In the free busy, availability, and consecutive availability endpoints of the Nylas Calendar API, there's a new field (`calendar`) that can be substituted for `emails` in the event that someone wants to check for availability/free-busy of other calendars within the same organization. 

# Usage
```javascript

// To check free-busy with calendars:
const freeBusy = nylas.calendars.freeBusy({
  startTime: 1590454800,
  endTime: 1590780800,
  calendars: [{
    accountId: "test_account_id",
    calendarIds: [
      "example_calendar_id_A",
       "example_calendar_id_B",
    ],
  }],
});

// To check availability with calendars:
const availability = nylas.calendars.availability({
  startTime: 1590454800,
  endTime: 1590780800,
  interval: 5,
  duration: 30,
  calendars: [{
    accountId: "test_account_id",
    calendarIds: [
      "example_calendar_id_A",
       "example_calendar_id_B",
    ],
  }],
});

// To check consecutive availability with calendars:
const availability = nylas.calendars.consecutiveAvailability({
  startTime: 1590454800,
  endTime: 1590780800,
  interval: 5,
  duration: 30,
  calendars: [{
    accountId: "test_account_id",
    calendarIds: [
      "example_calendar_id_A",
       "example_calendar_id_B",
    ],
  }],
});
```

# License


src/models/free-busy.ts
```diff
@@ -4,7 +4,34 @@ import Attributes from './attributes';
...
+  emails?: string[];
+  calendars?: FreeBusyCalendarsProperties[];
+};
+
+export type FreeBusyCalendarsProperties = {
+  accountId: string;
+  calendarIds: string[];
+};
+
+export class FreeBusyCalendars extends Model
```
mrashed-dev, line 16: I'm open to renaming this to something else, I'm just out of ideas!

src/models/calendar-restful-model-collection.ts
```diff
@@ -100,49 +101,39 @@ export default class CalendarRestfulModelCollection extends RestfulModelCollecti
...
-      for (const openHour of options.openHours) {
-        for (const email of openHour.emails) {
-          if (
-            !options.emails.some(row => row.includes(email)) &&
-            !freeBusyEmails.includes(email)
-          ) {
-            throw new Error(
-              'Open Hours cannot contain an email not present in the main email list or the free busy email list.'
-            );
+    if (!options.emails && !options.calendars) {
```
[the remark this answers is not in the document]
tbadun: 1. that looks right, you need at least one but both are fine
2. how detailed do you plan on going for error handling at this level?

src/models/calendar-restful-model-collection.ts
```diff
@@ -159,4 +150,32 @@ export default class CalendarRestfulModelCollection extends RestfulModelCollecti
...
+      duration_minutes: options.duration,
+      interval_minutes: options.interval,
+      start_time: options.startTime,
+      end_time: options.endTime,
+      buffer: options.buffer,
+      tentative_busy: options.tentativeBusy,
+      free_busy: freeBusyJson,
+      open_hours: openHoursJson,
+      calendars: calendarsJson,
+    };
```
tbadun, line 179: + `round_robin` and `event_collection_id` only for /availability (not consecutive or free-busy)

tbadun: there will be a few new additions to consecutive soon-ish so we'll keep you in the loop on those

tbadun: scratch that robin bit, just saw that you have it in `src/models/calendar-availability.ts`

src/models/calendar-restful-model-collection.ts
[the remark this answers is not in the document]
mrashed-dev: @tbadun I'm just implementing simple client-side rules that reflect what the API would see as an error, to prevent the user from having to make a call that will 100% fail. I don't want to add error checking for things that are dependant on other parts of the API account (for example, I don't want to check if an account has the Calendar scope set before making a Calendar-related call - that can be done by the API).

src/models/calendar-restful-model-collection.ts
mrashed-dev: @tbadun Awesome! I'll add in `event_collection_id`. Just a few questions to make sure I implement it properly:

1. What is the purpose of this field in terms of making an availability call?
2. Is it optional?
3. Is it a string?

Thank you!

tbadun: 1. A user can add a collection id to the metadata of one or more events with a given capacity. When looking for availability, the algorithm will grab all events in the specified collection and, if they are not currently at capacity, show those slots as available in addition to the regular availability data. This is mostly used by scheduler.
2. Yes
3. Also yes

src/models/calendar-restful-model-collection.ts
```diff
@@ -159,4 +139,48 @@ export default class CalendarRestfulModelCollection extends RestfulModelCollecti
...
+    // Instantiate objects from properties to get JSON formatted for the API call
+    const freeBusyJson = options.freeBusy
+      ? options.freeBusy.map(fb => new FreeBusy(fb).toJSON(true))
+      : [];
+    const openHoursJson = options.openHours
+      ? options.openHours.map(oh => new OpenHours(oh).toJSON(true))
+      : [];
+    const calendarsJson = options.calendars
+      ? options.calendars.map(cal => new FreeBusyCalendars(cal).toJSON(true))
+      : [];
```
james10424, line 155: These can all be written as the following:
```javascript
const freeBusyJson = options.freeBusy?.map(fb => new FreeBusy(fb).toJSON(true)) ?? [];
```

src/models/calendar-restful-model-collection.ts
```diff
@@ -159,4 +139,48 @@ export default class CalendarRestfulModelCollection extends RestfulModelCollecti
...
+      | SingleAvailabilityQuery
+      | ConsecutiveAvailabilityQuery
+  ): boolean {
+    if (
+      (!query.emails || query.emails.length == 0) &&
+      (!query.calendars || query.calendars.length == 0)
+    ) {
+      throw new Error("Must set either 'emails' or 'calendars' in the query.");
+    }
+    return true;
```
james10424, line 184: There is no need to return anything if the other option is to raise an error

src/models/free-busy.ts
```diff
@@ -4,7 +4,34 @@ import Attributes from './attributes';
...
+FreeBusyCalendars.attributes = {
+  accountId: Attributes.String({
+    modelKey: 'accountId',
+    jsonKey: 'account_id',
+  }),
+  calendarIds: Attributes.StringList({
+    modelKey: 'calendarIds',
+    jsonKey: 'calendar_ids',
+  }),
 };
```
james10424, line 35: It is probably cleaner to assign it inside the class definition.
```suggestion
export class FreeBusyCalendars extends Model
  implements FreeBusyCalendarsProperties {
  accountId = '';
  calendarIds: string[] = [];
  static attributes: Record<string, Attribute> = {
    accountId: Attributes.String({
      modelKey: 'accountId',
      jsonKey: 'account_id',
    }),
    calendarIds: Attributes.StringList({
      modelKey: 'calendarIds',
      jsonKey: 'calendar_ids',
    }),
  };

  constructor(props?: FreeBusyCalendarsProperties) {
    super();
    this.initAttributes(props);
  }
}
```

src/models/free-busy.ts
pooja169usp, line 16:
```suggestion
export class FreeBusyCalendar extends Model
```

src/models/free-busy.ts
```diff
@@ -4,7 +4,34 @@ import Attributes from './attributes';
 export type FreeBusyQuery = {
   startTime: number;
   endTime: number;
-  emails: string[];
+  emails?: string[];
+  calendars?: FreeBusyCalendarsProperties[];
+};
+
+export type FreeBusyCalendarsProperties = {
```
pooja169usp, line 11:
```suggestion
export type FreeBusyCalendarProperties = {
```

src/models/free-busy.ts
mrashed-dev: This is a really good idea. Thanks for picking this out! I'm not going to commit this change in this PR because _all_ classes are like this, so I'd rather make another PR where we change all the classes to be like this.

src/models/free-busy.ts
mrashed-dev: Good call out! Commit `04a3593` addresses this.

src/models/free-busy.ts
mrashed-dev: Good call out! Commit `04a3593` addresses this.

src/models/calendar-restful-model-collection.ts
```diff
@@ -159,4 +139,48 @@ export default class CalendarRestfulModelCollection extends RestfulModelCollecti
...
+      ? options.openHours.map(oh => new OpenHours(oh).toJSON(true))
+      : [];
+    const calendarsJson = options.calendars
+      ? options.calendars.map(cal => new FreeBusyCalendar(cal).toJSON(true))
+      : [];
+
+    return {
+      emails: options.emails,
+      duration_minutes: options.duration,
+      interval_minutes: options.interval,
```
james10424, line 160: Is there a reason why `minutes` is omitted? It would be better to stay consistent with the API

mrashed-dev: @james10424 that's a great callout, I'm not sure how it got missed during the refactor. Unfortunately if I were to make this change now, it would likely be considered a breaking change as people would need to adapt their code to the new keywords. Thus I wouldn't be able to make this change until the next major release. I've created a shortcut story to track this change when the time comes for the next major release though: https://app.shortcut.com/nylas/story/79594/update-availability-queries-to-match-api-notati

james10424 (approved):

PR #312, opened 2022-01-25
[79531] Move `attributes` and `collectionName` into classes as a static variables

mrashed-dev:
# Description
This PR's goal is for better readability, one of the proposed changes internally was to move all the `attributes` and `collectionName` properties into the class with as `static` properties.

# License


ozsivanov (approved):
LGTM - There is a lot of boilerplate code here though, not really sure if there's a great way to reduce it aside from pulling the attribute definitions outside of the class files and just importing them in.

AaronDDM (approved):

PR #318, opened 2022-02-02
[80399] Fix Virtual Calendar logic and serialization

mrashed-dev:
# Description
This PR fixes three issues in regards to Virtual Calendars:

1. Serializing a list of enums to JSON
2. The logic in determining a VirtualCalendar versus a NativeAuthentication
3. Unnecessary parameters during `Connect.authorize()` call leading to confusion

To expand on the third point, `Connect.authorize()` now takes 2-3 arguments compared to the previous 3-5:

1. name: string
2. emailAddress: string
3. clientId: string (optional, if not passed in it will just use the current clientId)

Compared to v6.0, `scope` and `settings` are no longer required as the API only accepts `['calendar']` for. scopes and an empty object for settings, so to prevent confusion we do not require these arguments and instead hardcode it in the class itself.

# License

### DIFF
--- a/src/models/connect.ts
+++ b/src/models/connect.ts
@@ -21,9 +21,7 @@ export enum Scope {
 export type VirtualCalendarProperties = {
   name: string;
   emailAddress: string;
-  scopes: Scope[];
   clientId?: string;
-  settings?: Record<string, any>;
 };
 
 export class VirtualCalendar extends Model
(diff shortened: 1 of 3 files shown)


src/models/connect.ts
```diff
@@ -166,7 +166,7 @@ export default class Connect {
     if (!auth.clientId) {
       auth.clientId = this.clientId;
     }
-    if (auth.hasOwnProperty('scopes')) {
+    if (auth.hasOwnProperty('provider') && (auth as any).provider != 'nylas') {
```
philrenaud, line 169: What's the meaning behind making sure the provider isn't nylas here? When might this come up?

philrenaud (approved):

src/models/connect.ts
mrashed-dev: @philrenaud I've received some feedback that this would be a nice to have given that some of our customers who migrated to v6.0 have missed this change of omitting provider.

PR #319, opened 2022-02-02
[79908] Add new field in `Draft` for adding files by file IDs

mrashed-dev:
# Description
This PR adds support for adding a new field, `fileIdsToAttach: string[]`, in `Draft` which will allow users who want to attach files to a draft via file IDs. Due to the existence of `fileIds()` function in the `Message` superclass, we can't set this variable to a friendlier name without causing a breaking change.

Please note that this variable only adds files to a draft, to remove a file it must be done on the files array. Upon saving we will make sure to merge unique values in the outgoing payload.

# Usage
```js

const draft = new Draft(nylas, {
  to: [{ name: 'Nylas Swag', email: 'swag@nylas.com' }],
  fileIdsToAttach = ["file_id_1", "file_id_2"],
});
draft.save();
```

# License

### DIFF
--- a/src/models/draft.ts
+++ b/src/models/draft.ts
@@ -18,6 +18,7 @@ export default class Draft extends Message implements DraftProperties {
   rawMime?: string;
   replyToMessageId?: string;
   version?: number;
+  fileIdsToAttach?: string[];
   static collectionName = 'drafts';
   static attributes: Record<string, Attribute> = {
     ...Message.attributes,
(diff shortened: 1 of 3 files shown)


src/models/draft.ts
```diff
@@ -39,12 +40,22 @@ export default class Draft extends Message implements DraftProperties {
     this.initAttributes(props);
   }
 
+  fileIds(): (string | undefined)[] {
+    let fileIds: (string | undefined)[] = super.fileIds();
+    if (this.fileIdsToAttach) {
+      fileIds = fileIds
+        .concat(this.fileIdsToAttach)
+        .filter((item, idx, mergedArray) => mergedArray.indexOf(item) === idx);
```
ozsivanov, line 48: Why do we need this filter?

mrashed-dev: To prevent duplication of file IDs. If a user saves a first version of a draft with two files of ids "file_1" and "file_2", then tries to update the draft and sets `this.fileIdsToAttach = ["file_2", "file_3"];` then without this filter the resulting array would be ["file_1", "file_2", "file_2", "file_3"] which I believe would result in an error on the API side.

ozsivanov: Gotcha, I think using a [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) would be a bit better for de-duping items. 
eg.
```
const deduplicatedArr = Array.from(new Set(arrayWithDuplicates));
```

mrashed-dev: @ozsivanov Beautiful! Done ✅

ozsivanov (approved):

Maksym-L (approved):

PR #324, opened 2022-02-08
[80753] Fix circular dependency issue in `Attribute`

mrashed-dev:
# Description
`RestfulModel` import statement causes a circular dependency issue because `RestfulModel` sets a static variable of `Attributes` type. Instead we should just cast `itemClass` as any object that extends `Model`.

# License

### DIFF
--- a/src/models/attributes.ts
+++ b/src/models/attributes.ts
@@ -1,10 +1,10 @@
 import Model from './model';
-import RestfulModel from './restful-model';
-
 // The Attribute class represents a single model attribute, like 'namespace_id'
 // Subclasses of Attribute like AttributeDateTime know how to covert between
 // the JSON representation of that type and the javascript representation.
 // The Attribute class also exposes convenience methods for generating Matchers.
+
+type AnyModel = (new (...args: any[]) => Model);
 
 export abstract class Attribute {
   modelKey: string;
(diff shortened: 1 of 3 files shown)


src/models/attributes.ts
```diff
@@ -72,7 +72,7 @@ class AttributeObject extends Attribute {
       return val;
     }
 
-    if (this.itemClass.prototype instanceof RestfulModel) {
+    if (this.itemClass.constructor.name == 'RestfulModel') {
```
ozsivanov, line 75: I'm not a big fan of this, but I see why you had to do it

ozsivanov (approved):

src/models/attributes.ts
mrashed-dev: @ozsivanov I'm not a big fan of it either, I'm actually working on an alternative since I assigned it to you, I'll message you when I have it!

mrashed-dev: @ozsivanov It's not _that_ much better but I think it's more accurate and it prevents scenarios like `Label` where its superclass is not `RestfulModel`, but `Folder` which extends `RestfulModel` (two classes removed).

ozsivanov (approved):

src/models/attributes.ts
```diff
@@ -1,11 +1,15 @@
 import Model from './model';
-import RestfulModel from './restful-model';
-
 // The Attribute class represents a single model attribute, like 'namespace_id'
 // Subclasses of Attribute like AttributeDateTime know how to covert between
 // the JSON representation of that type and the javascript representation.
 // The Attribute class also exposes convenience methods for generating Matchers.
 
+type AnyModel = new (...args: any[]) => Model;
+const isRestfulModel = (cls: any): boolean => {
```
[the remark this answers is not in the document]
mrashed-dev: I'll make this change, thank you! but `class` is a reserved keyword iirc so I'll keep it `cls`

PR #332, opened 2022-03-30
Add support for getting raw messages directly from the API

mrashed-dev:
# Description
This PR enables support for directly getting a raw message from the API. Currently on the SDK you would need to get a message first then run `getRaw()` on it, which results in 2 API calls. This function exists on `MessageRestfulModelCollection` allowing you to get the message as raw directly using a message ID resulting in just one API call.

# Usage
```ts

const messages = await nylas.messages.findRaw("message_id");
```

# License

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+### Unreleased
+* Add support for getting raw messages directly from the API
+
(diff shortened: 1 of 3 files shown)


arjunkalburgi (approved):
very simple and great code!

PR #334, opened 2022-03-31
Fix error when Nylas API returns non-JSON error

mrashed-dev:
# Description
Sometimes a call to the API might result in a 5xx server error with an HTML response. We should be accounting for this case and return a proper error instead of erroring out due to failed JSON parsing.

# License

### DIFF
--- a/src/nylas-connection.ts
+++ b/src/nylas-connection.ts
@@ -260,6 +260,13 @@ export default class NylasConnection {
                 error.serverError = body.server_error;
               }
               return reject(error);
+            }).catch(() => {
+              const error = new NylasApiError(
+                response.status,
+                "API Error",
+                "Error encountered during request, non-JSON formatted error received."
+              );
+              return reject(error);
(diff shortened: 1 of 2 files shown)


src/nylas-connection.ts
```diff
@@ -247,20 +247,30 @@ export default class NylasConnection {
...
+                }
+                return reject(error);
+              })
+              .catch(() => {
+                const error = new NylasApiError(
+                  response.status,
+                  'API Error',
+                  'Error encountered during request, non-JSON formatted error received.'
+                );
+                return reject(error);
```
krisgardiner, line 272: Should we be returning this error to the user? Should this parse the error on `response` if content is not json format, then return that error. Otherwise `response.json()` and return that error.

mrashed-dev: Great idea, just addressed!

krisgardiner (approved):

PR #338, opened 2022-04-08
Support Setting Redirect On Error in Authenticate URL Config

myleslinder:
Support the ability to set the `redirect_on_error` url param so that when using hosted authentication  the user can be redirected back to the app on error. 

# License

### DIFF
--- a/src/nylas.ts
+++ b/src/nylas.ts
@@ -183,8 +183,8 @@ class Nylas {
     if (options.provider != null) {
       url += `&provider=${options.provider}`;
     }
-    if(options.redirectOnError){
-      url += "&redirect_on_error=true"
+    if (options.redirectOnError) {
+      url += '&redirect_on_error=true';
     }
     return url;
   }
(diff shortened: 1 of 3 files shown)


src/nylas.ts
```diff
@@ -183,6 +183,9 @@ class Nylas {
     if (options.provider != null) {
       url += `&provider=${options.provider}`;
     }
+    if(options.redirectOnError){
+      url += "&redirect_on_error=true"
+    }
```
mrashed-dev, line 188: This block of code fails our `prettier` guideline check, the below suggestion should work!

```suggestion
    if (options.redirectOnError) {
      url += '&redirect_on_error=true';
    }
```
```

mrashed-dev (changes requested):
Thanks for your contribution, @myleslinder! Just made a comment above about styling on a line that's failing the CI check, if you commit that change we'll be good to merge!

mrashed-dev (approved):
Looks good to me! Thank you for your contribution @myleslinder 🚀

PR #339, opened 2022-04-12
Fix Outbox JSON Parsing Error 

myleslinder:
The outbox send method met a specific condition in `nylas-connection` where it would return undefined as opposed to the parsed json of the response. This would have the downstream base model throw an error when it tried to work with the json. It was meeting that condition because of an equality check with `response.size` and 0. [However with `node-fetch` `response.size` cannot be used/relied on](https://github.com/node-fetch/node-fetch/issues/1149). 

# License


mrashed-dev (approved):
Good catch @myleslinder, thank you for your contribution! 🚀

PR #340, opened 2022-04-12
Catch a JSON error and return undefined

mrashed-dev:
# Description
Sometimes the API won't respond with a JSON body, but given that the request completed without a status error, just return undefined.

# License

### DIFF
--- a/src/nylas-connection.ts
+++ b/src/nylas-connection.ts
@@ -248,6 +248,7 @@ export default class NylasConnection {
 
           if (response.status > 299) {
             return response
+              .clone()
               .json()
               .then(body => {
                 const error = new NylasApiError(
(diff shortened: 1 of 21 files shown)


src/nylas-connection.ts
```diff
@@ -309,7 +309,14 @@ export default class NylasConnection {
...
             } else {
-              return resolve(response.json());
+              return response
+                .json()
+                .then(json => {
+                  return resolve(json);
+                })
+                .catch(() => {
+                  return resolve(undefined);
+                });
```
AaronDDM, line 319: https://stevenklambert.com/writing/fetch-json-text-fallback/
This might not work!

mrashed-dev: Great find @AaronDDM, thank you! Fetch has been very frustrating to deal with regarding this.

james10424, line 319: nit:
```suggestion
                .then(resolve)
                .catch(() => resolve(undefined));
```

james10424 (approved):

src/nylas-connection.ts
mrashed-dev: @AaronDDM after doing some research looks like the article's right. just added in the new clone method. Thanks for spotting this!

AaronDDM (approved):
🎉

PR #341, opened 2022-04-12
[TSDK-277] Improve Outbox job status support

mrashed-dev:
# Description
This PR improves Outbox job status for the Node SDK by allowing for fetching an `OutboxJobStatus` object by job status ID, as well as add in some missing fields from an Outbox job status object.

# Usage
To get an outbox job status by ID, it's the same as you would get any Job Status, but in the background the SDK will identify if the job status is an outbox job status, and if so it will return the correct type:
```ts

const jobStatus = nylas.jobstatuses.find("outbox-job-status-id");
```

# License

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### Unreleased
 * Add support for revoking a single access token
+* Improve Outbox job status support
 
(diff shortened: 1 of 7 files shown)


pooja169usp (approved):
Looks good 👍🏽

PR #349, opened 2022-05-04
[Bugfix] Fix bug in CalendarRestfulModelCollection

crhayes:
I noticed the following error while attempting to use the "free busy" functionality:
image

Doing a manual CURL request with the same parameters was successful.

After some debugging I noticed that the code was attempting to `JSON.parse(json)`, where `json` was a JSON object and not a string. This resulted in the error I witnessed.

# License

### DIFF
--- a/__tests__/calendar-restful-model-collection-spec.js
+++ b/__tests__/calendar-restful-model-collection-spec.js
@@ -35,11 +35,9 @@ describe('CalendarRestfulModelCollection', () => {
         return Promise.resolve('body');
       },
       json: () => {
-        return Promise.resolve(
-          {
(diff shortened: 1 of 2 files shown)


crhayes:
@mrashed-dev Assuming there are no issues with this change, is it possible to have it merged, tagged, and released ASAP? It's causing a blocking issue for a use case we are exploring. Thanks!

mrashed-dev (approved):
Looks good, good catch @crhayes! Going to try fix the `prettier` offence.

crhayes:
Awesome, thanks. I should have run the linter 😅.

PR #350, opened 2022-05-04
Fix scheduler endpoint URL

mrashed-dev:
# Description
The scheduler endpoints were not using the correct base URL for API calls, resulting in errors for any scheduler operation. This PR ensures that the Scheduler endpoints use the correct url.

# License


tasdrake (approved):
LGTM

PR #352, opened 2022-05-05
[TSDK-296] Support collective and group events

mrashed-dev:
# Description
This PR bring support to [collective](https://developer.nylas.com/docs/connectivity/calendar/calendar-availability/#collective-meetings) and [group events](https://developer.nylas.com/docs/connectivity/calendar/calendar-availability/#group-meetings).

# License

### DIFF
--- a/src/models/event.ts
+++ b/src/models/event.ts
@@ -262,17 +262,7 @@ export default class Event extends RestfulModel {
   }
 
   save(params: {} | SaveCallback = {}, callback?: SaveCallback): Promise<this> {
-    if (
-      this.conferencing &&
-      this.conferencing.details &&
-      this.conferencing.autocreate
-    ) {
-      return Promise.reject(
-        new Error(
-          "Cannot set both 'details' and 'autocreate' in conferencing object."
-        )
-      );
-    }
+    this.validate();
     return super.save(params, callback);
   }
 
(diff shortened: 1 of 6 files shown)


src/models/event.ts
```diff
@@ -329,4 +336,26 @@ export default class Event extends RestfulModel {
...
+    ) {
+      throw new Error(
+        "Cannot set both 'details' and 'autocreate' in conferencing object."
+      );
+    }
+    if (
+      this.capacity &&
+      this.capacity != -1 &&
+      this.participants &&
+      this.participants.length <= this.capacity
```
ozsivanov, line 354: I don't think participants are included in the event capacity, since they're part of the meeting by default. cc @tbadun

tbadun: ^ that's just a weird internal verbiage thing, the participants list in an event is the sum total of organizers, participants, attendees, and invitees

tbadun: our cc-side validation looks like:
```python
def validate_capacity_request(self, capacity):
    # type: (int) -> None
    if (not isinstance(capacity, int)) or (
        (capacity == 0) or (capacity > 1000) or (capacity < -1)
    ):
        raise InputError(
            "Allowed capacity is between 1-1000 or -1 for unlimited: {}".format(
                capacity
            )
        )
    # only count none-organizer
    num_participants = len(
        [
            participant
            for participant in self.participants
            if participant.get("role") != "organizer"
        ]
    )

    if 0 < capacity < num_participants:
        raise InputError("Participants over capacity: {}".format(capacity))
```

ozsivanov: Cool, just wanted to double check 👍

ozsivanov (approved):
LGTM!

src/models/event.ts
tbadun: "participants" means nothing and everything at the same time, just to make communication around scheduler the most fun possible

mrashed-dev: @ozsivanov @tbadun thank you both!

PR #358, opened 2022-06-03
Remove use of `.clone` in `node-fetch` operations

mrashed-dev:
# Description
Since the release of v6.3.0 we incorporated the use of `node-fetch`'s `.clone` function which allows us to clone the response body before exhausting the stream, in case we run into some sort of parsing error (usually with JSON parsing). This led to some issues with fetching in general, and after some research it seems to be a known issue on node-fetch < 3.0.0 where if `.clone` is used on a large response, it won't properly read the stream (https://github.com/node-fetch/node-fetch/issues/396, https://github.com/node-fetch/node-fetch/issues/151). 

This issue was fixed in v3 of node-fetch but we're not able to directly jump to that version as it has breaking changes. As a result, we've switched to removing the use of `clone` and instead we just get the buffer as a string and try to parse the JSON ourselves, allowing us to still retain the buffered data in case of a parsing error. 

# License

### DIFF
--- a/src/nylas-connection.ts
+++ b/src/nylas-connection.ts
@@ -246,43 +246,30 @@ export default class NylasConnection {
           }
 
           if (response.status > 299) {
-            return response
-              .clone()
-              .json()
-              .then(body => {
+            return response.text().then(body => {
+              try {
+                const jsonError = JSON.parse(body);
                 const error = new NylasApiError(
                   response.status,
-                  body.type,
-                  body.message
+                  jsonError.type,
+                  jsonError.message
                 );
-                if (body.missing_fields) {
-                  error.missingFields = body.missing_fields;
+                if (jsonError.missing_fields) {
+                  error.missingFields = jsonError.missing_fields;
                 }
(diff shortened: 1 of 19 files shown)


src/nylas-connection.ts
```diff
@@ -246,43 +246,30 @@ export default class NylasConnection {
           }
 
           if (response.status > 299) {
-            return response
-              .clone()
-              .json()
-              .then(body => {
+            return response.text().then(body => {
+              try {
+                const jsonError = JSON.parse(body);
```
ozsivanov, line 251: So `jsonError` here would just be the parsed body, why do we expect it to be an error?
`JSON.parse()` throws an exception if there is an error with the provided JSON, so the catch would be handling that case already.

ozsivanov: Oh I misunderstood the variable naming, this is just the parsed error object because the response error code was `> 299`. Maybe consider renamind this variable.

ozsivanov (approved):
LGTM

PR #363, opened 2022-06-10
Support Nylas API v3.0

isaacnass:
- Bump version (`6.4.1` => `6.4.2`)
- Add `authState` object support/typing to both `account` and `management-account` endpoints
- Mark `syncState` as deprecated, as it is now superceded by `authState`

### DIFF
--- a/src/models/account.ts
+++ b/src/models/account.ts
@@ -1,13 +1,30 @@
 import RestfulModel from './restful-model';
 import Attributes, { Attribute } from './attributes';
 import NylasConnection from '../nylas-connection';
+import { NativeAuthenticationProvider, Scope } from './connect';
+
+export enum AuthStatusState {
+  Valid = 'valid',
+  Invalid = 'invalid',
+}
+
+export type AuthStatusProperties = {
+  state: AuthStatusState;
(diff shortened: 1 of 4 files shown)


src/models/account.ts
```diff
@@ -1,13 +1,30 @@
...
+  error_message?: string;
+};
 
 export type AccountProperties = {
   name: string;
   emailAddress: string;
   provider: string;
   organizationUnit: string;
+  /**
+   * @deprecated will be removed in future versions. use `authStatus` instead
```
johnjjung, line 24: This is actually not finalized. SyncState will likely not be deprecated but mapped to AuthStatus, and drop things that don't make a lot of sense like `partial` and a few others. AuthStatus would be an additional attribute. The reason is this a major breaking change would require so many downstream changes for our customers that the ROI to introduce a breaking change is not worth it.

johnjjung:
Likely we will keep backwards compatibility with API 2.6 with the introduction of AuthStates

mrashed-dev:
API v3 is going to be a bigger undertaking so closing this PR as plans have changed since then. Thanks @isaacnass for the contribution :)

PR #366, opened 2022-06-16
[Isaac] v7 Tunneling Service

isaacnass:
- Create a new tunneling service that mimics the functionality of the CLI
- Intended use is to add `websocketEventListener` (or something like that) option to the upcoming server bindings feature, that automatically calls this
- Events should be output using the same mechanism as the automatic webhook registry
- Add the `message.bounce` webhook event (as it's in the CLI)
- Add region configs, carried over from the CLI

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -23,6 +23,8 @@
         "@types/backoff": "^2.5.1",
         "@types/jest": "^25.1.4",
         "@types/node-fetch": "^2.5.8",
+        "@types/uuid": "^8.3.4",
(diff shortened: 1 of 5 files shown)


mrashed-dev:
Tagging in the CLI master himself, @AaronDDM. Thanks @isaacnass!

AaronDDM:
That was quick! Awesome :) this resolves the Nylas CLI dependency issue for node.

AaronDDM (approved):
This is great

mrashed-dev (approved):
Looks good to me!

PR #367, opened 2022-06-16
Add binding server frameworks and Express.js binding support

mrashed-dev:
# Description
This PR enables the Nylas SDK to build middleware to integrate into your backend application with routes to some endpoints that would help with integrating authentication and webhook event. Currently provided routes:
* POST `/nylas/webhook` - Endpoint that expects webhook responses from and emits the decoded object. Verifies the Nylas webhook signature and emits a `WebhookDelta` object as a `WebhookTriggers` type
* POST `/nylas/generate-auth-url` - Endpoint that takes an optional email address and success redirect path, returns a URL used for authentication (from `Nylas.urlForAuthentication`)
* POST `/nylas/exchange-mailbox-token` - Endpoint that takes a token from the JSON and exchanges that code for token (using `Nylas.exchangeCodeForToken`). Returns 200 on success. 

# Usage
Here's an example application that connects an express backend to a Nylas SDK instance and handles a token exchange event and a webhook trigger (account.connected):
```js
const express = require('express');
const { WebhookTriggers } = require('nylas/lib/models/webhook');
const { ServerEvents } = require('nylas/lib/server-bindings/server-binding');
const { Scope } = require('nylas/lib/models/connect');
const { ServerBindings } = require('nylas/lib/config');

const port = 9000;
// The uri for the frontend

  clientId: "CLIENT_ID",
  clientSecret: "CLIENT_SECRET",
});
const app = express();
let accessToken = '';

// Use the express bindings provided by the SDK and pass in additional configuration such as auth scopes
const expressBinding = new ServerBindings.express(nylasClient, {
  defaultScopes: [
    Scope.EmailModify,
    Scope.EmailSend,
    Scope.ContactsReadOnly,
  ],
  clientUri,
});

// Handle when a token has been exchanged
expressBinding.on(ServerEvents.TokenExchange, ({ accessTokenObj }) => {
  // Normally store the access token in the DB
  accessToken = accessTokenObj.accessToken;
  console.log('Access Token was generated for: ' + accessTokenObj.emailAddress);
});

// Handle a specific trigger, like account.connected
expressBinding.on(WebhookTriggers.AccountConnected, (payload) => {
  console.log("Webhook trigger received, account connected. Data: " + payload);
});

// Mount the express middleware to your express app
const nylasMiddleware = expressBinding.buildMiddleware();
app.use('/', nylasMiddleware);

// Start the Nylas webhook
expressBinding.startDevelopmentWebsocket();

// Add some custom paths
app.get('/', (req, res) => res.status(200).send('Hello World!'));

app.listen(port, () => console.log('App listening on port ' + port));
```

# License


AaronDDM:
@mrashed-dev This is amazing! 
One thing, is it possible to make the routing injection follow the common standard for express middleware apps?
https://expressjs.com/en/guide/using-middleware.html#middleware.router

Essentially you are giving people the flexibility to set their own binding route (e.g. `/dev/...` or something like that)

src/server-bindings/server-binding.ts
```diff
@@ -0,0 +1,84 @@
...
+import AccessToken from '../models/access-token';
+import Nylas from '../nylas';
+
+type SupportedFrameworks = Express;
+
+export enum ServerEvents {
+  TokenExchange = 'token-exchange',
+}
+
+export type ServerBindingOptions = {
```
ozsivanov, line 16: nit - this should be an `interface`

ozsivanov (approved):
LGTM! +1 for @AaronDDM's recommendation of implementing this as a middleware

src/server-bindings/server-binding.ts
```diff
@@ -0,0 +1,130 @@
...
+    rawBody: Buffer
+  ): boolean {
+    const digest = crypto
+      .createHmac('sha256', secret)
+      .update(rawBody)
+      .digest('hex');
+    return digest === xNylasSignature;
+  }
+
+  startDevelopmentWebsocket(): void {
```
isaacnass, line 85: This is a bit confusing that it has to be called manually. Would recommend running this automatically on mount if `useDevelopmentWebsocketEventListener` is truthy

AaronDDM (approved):

src/server-bindings/server-binding.ts
mrashed-dev: We removed mount in favour of this becoming middleware. As a result though we've moved out the config to the function parameter.

PR #373, opened 2022-06-21
Remove `routePrefix` config from server bindings

mrashed-dev:
# Description
This PR removes confusion as behind the scenes, `buildRoute` was taking either `routPrefix` or `/nylas` (if unset) and appending it to the paths. So if someone mounts the middleware at `/nylas`, the routes would have been `/nylas/nylas/...`. 

# Usage
If we want to mount endpoints to /nylas:
```js
const express = require('express');
const { ServerBindings } = require('nylas/lib/config');

const port = 9000;
// The uri for the frontend

    clientId: clientId,
    clientSecret: clientSecret,
});
const app = express();
app.listen(port, () => console.log('App listening on port ' + port));

// Use the express bindings provided by the SDK and pass in additional configuration such as auth scopes
const nylasMiddleware = new ServerBindings
    .express(nylasClient, {
        defaultScopes: [
            Scope.EmailModify,
            Scope.EmailSend,
        ],
        clientUri,
    })
    .buildMiddleware();

// Mount it on /nylas
app.use('/nylas', nylasMiddleware);
```

# License

### DIFF
--- a/src/server-bindings/server-binding.ts
+++ b/src/server-bindings/server-binding.ts
@@ -19,7 +19,6 @@ export enum ServerEvents {
 
 export type ServerBindingOptions = {
   defaultScopes: Scope[];
-  routePrefix?: string;
   clientUri?: string;
 };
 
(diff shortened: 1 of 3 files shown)


AaronDDM (approved):

src/server-bindings/server-binding.ts
```diff
@@ -30,7 +30,6 @@ export abstract class ServerBinding extends EventEmitter
   routePrefix?: string;
```
isaacnass, line 30: Dont forget this

mrashed-dev: Good catch!

PR #381, opened 2022-07-25
Fix typing issues in event and scheduler open hours

mrashed-dev:
# Description
This PR fixes issues relating to typing in:
- `Event`: `originalStartTime` should deserialize to `Date`, so replace `number` type to `Date`.
- `SchedulerBookingOpenHours`: The `account_id` filed from the API was being deseralized to `dropdownOptions` instead of `accountId`.

# License

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,8 @@
 
 ### Unreleased
 * Add `interval_minutes` field in Scheduler booking config
-* Adjust `Event.originalStartTime` type to include `Date`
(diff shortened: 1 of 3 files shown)


AaronDDM (approved):
@mrashed-dev Does the `originStartTime` ever return a number? If not, isn't it already broken, so no harm in getting rid of the `number` type?

mrashed-dev:
@AaronDDM it deserializes to a `Date` type so you might be right...

PR #382, opened 2022-07-28
Abstracting middleware logic

mrashed-dev:
# Description
This PR's main change is pulling out the underlying logic of the routes used in the middleware created for Express and moving it to `Routes`. This new module's purpose is to allow implementing clients that don't use one of the supported frameworks served by `ServerBindings` still an easy way to integrate Nylas into their code and pair it with our frontend SDKs we provide (@nylas/nylas-js, @nylas/nylas-react).

This code also brings forward a breaking change, the `Nylas.exchangeCodeForToken` now expects the optional callback parameter to anticipate a `AccessToken` object instead of a string for the `accessToken` parameter. 

# Usage
We're going to use `fastify` as we currently do not support this framework to demonstrate how using `Routes` can make  it easy to implement these routes:
```js
const fastify = require('fastify')
const cors = require('@fastify/cors')

const { v4: uuidv4 } = require("uuid");
const { mockDb } = require("./utils/mock-db");

const { Scope } = require('nylas/lib/models/connect');
const { Routes: NylasRoutes }  = require('nylas/lib/services/routes');
const { DefaultPaths } = require('nylas/lib/services/routes')

const port = 9000;

// The uri for the frontend

// Nylas application credentials
const clientId = '$CLIENT_ID';
const clientSecret = '$CLIENT_SECRET';

    clientId: clientId,
    clientSecret: clientSecret,
});

// Use the routes provided by the Nylas Node SDK to quickly implement the authentication flow
const { buildAuthUrl, exchangeCodeForToken } = NylasRoutes(nylasClient);

// Init the fastify server
const app = fastify();

// Enable CORS
app.register(cors)

// Configure the Nylas routes using your flavour of backend framework, fastify in this case
app.post(DefaultPaths.buildAuthUrl, async (req, res) => {
  const authUrl = await buildAuthUrl({
    scopes: [
      Scope.EmailReadOnly,
    ],
    emailAddress: req.body.email_address,
    successUrl: req.body.success_url,
    clientUri,
  });
  res.status(200).send(authUrl);
});

app.post(DefaultPaths.exchangeCodeForToken, async (req, res) => {
  try {
    const { accessToken, emailAddress } = await exchangeCodeForToken(req.body.token);

    // Replace this mock code with your actual database operations
    const user = await mockDb.createOrUpdateUser(emailAddress, {
      accessToken,
      emailAddress,
    })
  
    // Return an authorization object to the user
    res.send({
      id: user.id,
      emailAddress: user.emailAddress,
    });
  } catch (e) {
    res.status(500).send((e).message);
  }
});

// Add some routes for the backend
app.get('/', (req, res) => res.status(200).send('Ok'));

// Start listening on port 9000
app.listen({ port }).then(() => console.log('App listening on port ' + port));
```

# License

### DIFF
--- a/src/services/routes.ts
+++ b/src/services/routes.ts
@@ -1,6 +1,7 @@
 import { Scope } from '../models/connect';
 import Nylas from '../nylas';
 import AccessToken from '../models/access-token';
+import crypto from 'crypto';
 
 export enum DefaultRoutes {
   buildAuthUrl = '/generate-auth-url',
@@ -24,6 +25,11 @@ type Routes = {
     nylasClient: Nylas,
     code: string
   ) => Promise<AccessToken>;
+  verifyWebhookSignature: (
+    nylasClient: Nylas,
+    nylasSignature: string,
+    rawBody: Buffer
+  ) => boolean;
 };
 
 export const Routes = (): Routes => {
@@ -48,7 +54,25 @@ export const Routes = (): Routes => {
     return await nylasClient.exchangeCodeForToken(code);
   };
 
-  return { buildAuthUrl, exchangeCodeForToken };
+  /**
+   * Verify incoming webhook signature came from Nylas
+   * @param nylasSignature The signature to verify
+   * @param rawBody The raw body from the payload
+   * @return true if the webhook signature was verified from Nylas
+   */
+  const verifyWebhookSignature = (
+    nylasClient: Nylas,
+    nylasSignature: string,
+    rawBody: Buffer
+  ): boolean => {
+    const digest = crypto
+      .createHmac('sha256', nylasClient.clientSecret)
+      .update(rawBody)
+      .digest('hex');
+    return digest === nylasSignature;
+  };
+
+  return { buildAuthUrl, exchangeCodeForToken, verifyWebhookSignature };
 };
 
 export default Routes;
--- a/src/services/routes.ts
+++ b/src/services/routes.ts
@@ -4,10 +4,10 @@ import AccessToken from '../models/access-token';
 import crypto from 'crypto';
 import { ExchangeCodeForTokenCallback } from '../config';
 
-  buildAuthUrl = '/generate-auth-url',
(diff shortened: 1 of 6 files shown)


src/services/routes.ts
```diff
@@ -0,0 +1,90 @@
+import { Scope } from '../models/connect';
+import Nylas from '../nylas';
+import AccessToken from '../models/access-token';
+import crypto from 'crypto';
+import { ExchangeCodeForTokenCallback } from '../config';
+
+export enum DefaultRoutes {
+  buildAuthUrl = '/generate-auth-url',
+  exchangeCodeForToken = '/exchange-mailbox-token',
+  webhooks = '/webhook',
```
krisgardiner, line 10: What do you think of adding the `/nylas` root route here to reduce the risk of route conflicts on a user's application?
```suggestion
  buildAuthUrl = '/nylas/generate-auth-url',
  exchangeCodeForToken = '/nylas/exchange-mailbox-token',
  webhooks = '/nylas/webhook',
```

krisgardiner: Implementation would be `app.post(DefaultRoutes.buildAuthUrl, async (req, res) => {` instead.

Unless we don't want to be that opinionated about this, but some pros to reserving a "root" level route.

krisgardiner (approved):

src/services/routes.ts
AaronDDM: I think it's fine being that opinionated, because in that case they can just not use these values right?

AaronDDM: You guys think it is confusing to have `Routes` and `DefaultRoutes`? Maybe `DefaultPaths` or something instead?

mrashed-dev: Sounds good:

- Defaulting to `/nylas` for paths as @krisgardiner suggested
- Allow overriding the paths in the ServerBinding config
- Renamed `DefaultRoutes` to `DefaultPaths` as @AaronDDM suggested

krisgardiner (approved):

PR #387, opened 2022-08-26
Add additional Event fields

mrashed-dev:
# Description
Add the following fields to `Event`:
* customeEventId: string
* oragnizerEmail: string (read only)
* organizerName: string (read only)
* visibility: string (read only)

# License


theblueskies:
LGTM

PR #399, opened 2022-10-12
TSDK-417 Adding Event Reminders to Event Object

kraju3:
# Description

Adding Event Reminder method to the Event object. Just creating a PR to see if I am on the right track here. I can see that our Object Schema from our API and the response we get back differ. There is some duplication that needed to be done to get this working. However, in this.validate() I am adding some validations to make sure the request passes through.

Questions And Concerns

So I am making sure no reminders are set on PUT calls using the id field
I still need to add some tests to test out read and write for this property
Also when you get the response back we get a reminders object :( So Mapping is different, so I had to duplicate some value with how our fromJson and toJson behave. But the weird thing is reminder_minutes is a string in the post request but it is an integer in the response

# License

### DIFF
--- a/src/models/event.ts
+++ b/src/models/event.ts
@@ -11,6 +11,11 @@ import NylasConnection from '../nylas-connection';
 import EventNotification, {
   EventNotificationProperties,
 } from './event-notification';
+import {
+  EventReminder,
+  EventReminderMethod,
+  EventReminderProperties,
+} from './event-reminder-method';
 
 export enum ICSMethod {
   Request = 'request',
@@ -57,6 +62,9 @@ export type EventProperties = {
   organizerName?: string;
   visibility?: string;
   customerEventId?: string;
+  reminderMinutes?: string;
+  reminderMethod?: EventReminderMethod;
+  reminders?: EventReminderProperties;
 };
 
 export default class Event extends RestfulModel {
@@ -81,7 +89,9 @@ export default class Event extends RestfulModel {
   originalStartTime?: Date;
   capacity?: number;
   conferencing?: EventConferencing;
+  readonly reminders?: EventReminder;
   reminderMinutes?: string;
+  reminderMethod?: EventReminderMethod;
   notifications?: EventNotification[];
   roundRobinOrder?: string[];
   metadata?: object;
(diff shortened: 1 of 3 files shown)


abdullahtariq1171:
Thank you [kraju3](https://github.com/kraju3) for raising the PR.

src/models/event.ts
```diff
@@ -394,5 +413,17 @@ export default class Event extends RestfulModel {
         'The number of participants in the event exceeds the set capacity.'
       );
     }
+
+    if (
+      this.id &&
+      ((this.reminders &&
+        (this.reminders.reminderMethod || this.reminders.reminderMinutes)) ||
```
mrashed-dev, line 420: Why do we throw an error if the event has a reminders object set? In this case the user can never update an event that was created with a reminder as ID will be set and `this.reminders` will be set.

src/models/event.ts
```diff
@@ -167,6 +177,15 @@ export default class Event extends RestfulModel {
...
     }),
+    reminderMethod: Attributes.String({
+      modelKey: 'reminderMethod',
+      jsonKey: 'reminder_method',
+    }),
+    reminders: Attributes.Object({
+      modelKey: 'reminders',
+      jsonKey: 'reminders',
+      itemClass: EventReminder,
+    }),
```
mrashed-dev, line 188: I think we shouldn't throw an error if the reminders object exists in an event, but make it read only so we don't serialize it and invoke an API error.

```suggestion
    reminders: Attributes.Object({
      modelKey: 'reminders',
      jsonKey: 'reminders',
      itemClass: EventReminder,
      readOnly: true
    }),
```

src/models/event.ts
```diff
@@ -394,5 +413,17 @@ export default class Event extends RestfulModel {
...
       );
     }
+
+    if (
+      this.id &&
+      ((this.reminders &&
+        (this.reminders.reminderMethod || this.reminders.reminderMinutes)) ||
+        this.reminderMethod ||
+        this.reminderMinutes)
+    ) {
```
mrashed-dev, line 424: After we make the reminders object readonly we can omit it from this check.

```suggestion
    if (
      this.id &&
        this.reminderMethod ||
        this.reminderMinutes)
    ) {
```

mrashed-dev (approved):
Looks good to me! Will add additional test coverage.

mrashed-dev (approved):

PR #414, opened 2022-12-16
[v7] Introduce query params, improved typing, other DevEx improvements for

mrashed-dev:
# Description
### ⚠️ Still a WIP

This PR holds new changes coming to the Node SDK v7 where we aim to further improve the developer experience of using restful model collections in this SDK. Currently this PR specifically brings:
* Query parameter support. Now we have a reference of all the query parameters each collection supports, eliminating the need to check the API Reference docs and reduces the chance of errors using unsupported parameters.
* Improved typing. Reduced the reliance on generics as generics don't translate to good type hinting in IDEs (they end up being `any` which we'd like to avoid), so we have started re-defining methods that use generics such as `list` with specific types.
* Reducing import statements. The number one feedback we got from v6 was that import statements can be a little confusing and hard to track. So we've introduced a new `create` method for all collections that support creating new resources, that eliminate the need for importing the parent class as well as it saves it to the API without needing an extra step. Creating a new object in the API can now be reduced to one line with no additional imports.

# License

### DIFF
--- a/src/models/calendar-restful-model-collection.ts
+++ b/src/models/calendar-restful-model-collection.ts
@@ -8,7 +8,6 @@ import CalendarAvailability, {
   OpenHours,
   SingleAvailabilityQuery,
 } from './calendar-availability';
-import { SaveCallback } from './restful-model';
 import { MetadataQuery, RestfulQuery } from './model-collection';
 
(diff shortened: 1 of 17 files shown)


mrashed-dev:
No longer relevant with the new direction we are going in.

PR #417, opened 2023-01-09
Add new attributes for Event webhook notifications

mrashed-dev:
# Description
This PR adds two new fields and expands the type of the metadata in WebhookNotification for `Event` objects. The following changes occurred:
- Added `calendarId` to `WebhookObjectAttributes`
- Added `createdBeforeAccountConnection` to `WebhookObjectAttributes`
- Expanded metadata to support `Record<string, unknown>`, will deserialize to `MessageTrackingDataProperties` if it's a message or thread object as it did before.

# License

### DIFF
--- a/src/models/webhook-notification.ts
+++ b/src/models/webhook-notification.ts
@@ -238,7 +238,7 @@ export type WebhookObjectDataProperties = {
   accountId: string;
   namespaceId: string;
   object: string;
-  metadata?: MessageTrackingDataProperties | Record<string, object>;
+  metadata?: MessageTrackingDataProperties | Record<string, unknown>;
   objectAttributes?: WebhookObjectAttributesProperties;
 };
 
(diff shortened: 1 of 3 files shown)


krisgardiner (approved):

src/models/webhook-notification.ts
```diff
@@ -271,6 +282,27 @@ export class WebhookObjectData extends Model
   constructor(props?: WebhookObjectDataProperties) {
     super();
     this.initAttributes(props);
+    if (
+      this.metadata &&
+      (this.object === 'message' || this.object === 'thread')
```
ozsivanov, line 287: Not a blocker, but I'm wondering if it'll make sense to create a string enum for handling object types  in the future.

ozsivanov (approved):
LGTM

src/models/webhook-notification.ts
mrashed-dev: @ozsivanov that's a good shout, I'll merge it in as is and open a new ticket to make a string enum, thank you 😄

PR #424, opened 2023-02-21
Update restful-model-collection.ts

philiiiiiipp:
Forward the generic type to allow for better type safety in the child models.

I assume this was just overlooked.

# License


mrashed-dev (approved):
An oversight indeed. Thanks @philiiiiiipp for catching that!

Issue #421, opened 2023-01-30
Allow a timeout for http requests


philiiiiiipp:
**Is your feature request related to a problem? Please describe.**
We see that some requests are taking forever to complete (more than 900s) because of the underlying email server timing out. In our case we use aws lambda to send emails which with a maximum runtime of 900s is not happy.

**Describe the solution you'd like**
We would like to hand a timeout to the nylas sdk which gets forwarded to `node-fetch` so we can handle long running calls appropriately.

I can see this to work either globally via the `NylasConnection` or the `NylasConfig` doing something like
```
import Nylas from 'nylas';

Nylas.config({
  clientId: '',
  clientSecret: '',
  httpOptions: {
    timeout: 3000,
    connectTimeout: 1000,
  },
});
```

or via the individual calls to `send` etc. I see the option via the `.config` easier because the sdk allows for a callback and adding some options to it would become a bit cumbersome.

mrashed-dev:
@philiiiiiipp Thanks for opening this issue. We will take a look at implementing this feature and report back with updates. Thanks!

mrashed-dev:
Hey @philiiiiiipp, we are looking into adding this feature for the Node SDK - can you expand on the difference between `timeout` and `connectTimeout`? Thanks!

philiiiiiipp:
I copied this from the `aws-sdk` I think there they distinguish between a connectTimeout e.g. ssl handshake etc. and the actual call timeout. They also offer mechanisms to retry based on those timeouts.

However, I feel for the `nylas-sdk` that would not be necessary, also given that `node-fetch` does not easily allow you to do this as far as I know. 

For us a `timeout` option would already be a massive improvement.

mrashed-dev:
@philiiiiiipp thank you for your response and your (separate) PR, will look into adding this feature now and will review & merge your PR as well. Thanks for your contributions!

mrashed-dev:
@philiiiiiipp The feature has been implemented and should be part of tomorrow's release! Thanks for your patience. To use the timeout feature, pass in the extra `timeout` configuration option:

```js
const Nylas = require('nylas');


// Initialize an instance of the Nylas SDK using the client credentials and timeout
Nylas.config({
  clientId: "CLIENT_ID",
  clientSecret: "CLIENT_SECRET",
  timeout: 60 // 60 second timeout
});
```

PR #427, opened 2023-03-09
Allow configurable timeout for API calls

mrashed-dev:
# Description
This PR enables support for configuring a timeout for all outgoing API calls, closing #421.

# Usage
To set the timeout you can add it as part of the initial Nylas SDK config:
```js


// Initialize an instance of the Nylas SDK using the client credentials and timeout
Nylas.config({
  clientId: "CLIENT_ID",
  clientSecret: "CLIENT_SECRET",
  timeout: 60 // 60 second timeout
});
```

# License

### DIFF
--- a/src/nylas-connection.ts
+++ b/src/nylas-connection.ts
@@ -332,7 +332,7 @@ export default class NylasConnection {
           console.error(`Error encountered during request:\n${err.stack}`);
           return reject(err);
         })
-        .finally(() => {
+        .then(() => {
           clearTimeout(timeout);
         });
     });
--- a/src/nylas-connection.ts
+++ b/src/nylas-connection.ts
@@ -227,8 +227,9 @@ export default class NylasConnection {
 
   request(options: RequestOptions): Promise<any> {
     const req = this.newRequest(options);
+    const timeout = config.timeout;
     return new Promise<any>((resolve, reject) => {
-      return fetch(req)
+      return fetch(req, { timeout })
         .then(response => {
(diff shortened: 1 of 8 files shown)


src/nylas-connection.ts
```diff
@@ -316,8 +329,15 @@ export default class NylasConnection {
...
         })
         .catch((err: Error) => {
+          if (err && err.name && err.name === 'AbortError') {
+            console.warn('Request timed out');
+            return reject(err);
+          }
           console.error(`Error encountered during request:\n${err.stack}`);
           return reject(err);
+        })
+        .then(() => {
```
AaronDDM, line 339:
```suggestion
        .finally(() => {
```

This is to clear the timeout regardless of what happens, right (e.g. success or timeout)?

AaronDDM (approved):

src/nylas-connection.ts
```diff
@@ -227,8 +230,18 @@ export default class NylasConnection {
...
+    if (config.timeout && config.timeout !== 0) {
+      controller = new AbortController();
+      timeout = setTimeout(() => {
+        controller.abort();
+      }, 150);
+    }
+
     return new Promise<any>((resolve, reject) => {
-      return fetch(req)
+      return fetch(req, { signal: controller?.signal as AbortSignal })
```
krisgardiner, line 244: This should prevent you from type casting.
```suggestion
      return fetch(req, { ...(controller && { signal: controller.signal }) })
```

PR #429, opened 2023-03-13
Add `startTimezone`, `endTimezone` and `timezone` fields to `When`

Fylax:
As specified in [Nylas Docs](https://developer.nylas.com/docs/api/#tag--Events--event-subobjects), the `When` object within a `Event` API support the following fields:
- `start_timezone`
- `end_timezone`
- `timezone`

This PR adds the related fields to the `When` object

# License


mrashed-dev (approved):
Thank you for your contribution @Fylax!

PR #430, opened 2023-03-13
Add HTTP status 409 to error mapping

johndiazyembo:
We have occasionally seen API responses with an HTTP status of 409. That status is not mapped and causes an exception. This PR adds that mapping to correct that.

Additionally, [the API error documentation](https://developer.nylas.com/docs/api/errors) should also be updated to reflect the possibility of a 409 error.

# License


mrashed-dev:
Thanks for your contribution @johndiazyembo, will forward your feedback to our docs team!

mrashed-dev (approved):

PR #439, opened 2023-03-24
Support authentication grant models and CRUD operations

mrashed-dev:
# Description
- Added grant models
- Added grant CRUD operations
- Abstracted resource methods and moved them into baseResource
- Added update PATCH operation
- Fixed method being discarded before outgoing calls

# License

### DIFF
--- a/src/resources/baseResource.ts
+++ b/src/resources/baseResource.ts
@@ -1,4 +1,3 @@
-/* eslint-disable no-console */
 import { ZodType } from 'zod';
 import APIClient from '../apiClient';
 import { OverridableNylasConfig } from '../config';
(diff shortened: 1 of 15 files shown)


b3ndoi (approved):
LGTM 😄

src/resources/baseResource.ts
```diff
@@ -113,14 +136,73 @@ export class BaseResource {
...
+        path,
+        overrides,
+      },
+      {
+        responseSchema,
+      }
+    );
+  }
+
+  private payloadRequest<T>(
```
agtang96, line 164: i think this has a very small chance of ending up being too catch all but we'll see

agtang96 (approved):

PR #440, opened 2023-03-24
Hosted auth

b3ndoi:
Auth methods for SDK for generating hosted auth links (`code` or `adminconsent`), code exchange handling, token exchange handling & revoking tokens
# License

### DIFF
--- a/src/schema/auth.ts
+++ b/src/schema/auth.ts
@@ -1,29 +1,26 @@
 import { z } from 'zod';
 
-const Providers = z.union([
-  z.literal('google'),
-  z.literal('microsoft'),
-]).optional();
-const AccessType = z.union([
-  z.literal('online'),
-  z.literal('offline'),
-]);
-const ResponseType = z.union([
-  z.literal('code'),
-  z.literal('adminconsent'),
-]);
-const Prompt = z.union([
-  z.literal('select_provider'),
-  z.literal('detect'),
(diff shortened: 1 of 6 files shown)


src/schema/auth.ts
```diff
@@ -0,0 +1,31 @@
+type AccessType = 'online' | 'offline';
+type Provider = 'google' | 'microsoft';
```
agtang96, line 2: is yahoo missing here?

src/schema/auth.ts
```diff
@@ -0,0 +1,31 @@
...
+export type AdminConsentAuth = SharedAuthParams & {
+  credentialId: string;
+};
+
+type SharedAuthParams = {
+  accessType?: AccessType;
+  prompt?: string;
+  redirectUri: string;
+  scope?: Array<string>;
+  includeGrantScopes: boolean;
```
agtang96, line 13: i don't see this in the UAS docs o.o

agtang96 (approved):

src/schema/auth.ts
b3ndoi: IDK if we support Yahoo for other stuff I know passthru does not

src/schema/auth.ts
b3ndoi: It's still inside the codebase so maybe they forgot to add it to the docs 🤔

PR #444, opened 2023-04-04
Added token validation

b3ndoi:
Added token validation for issued access tokens & also ID tokens, both of them is successful return an OpenID for the corresponding user, also needed to extend error handling because of new struct of error on that endpoint

# License


src/index.js
mrashed-dev, line 1: @b3ndoi I think this was accidentally pushed, might wanna delete!

PR #451, opened 2023-04-10
Added hosted auth

b3ndoi:
# License

### DIFF
--- a/src/resources/auth.ts
+++ b/src/resources/auth.ts
@@ -6,7 +6,9 @@ import {
   AdminConsentAuth,
   CodeExchangeRequest,
   TokenExchangeRequest,
-  HostedAuthRequest
+  HostedAuthRequest,
+  HostedAuth,
+  HostedAuthSchema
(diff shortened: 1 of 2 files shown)


src/resources/auth.ts
```diff
@@ -162,6 +163,26 @@ export class Auth extends BaseResource {
...
+   */
+  public async hostedAuth(payload: HostedAuthRequest): Promise<boolean> {
+    await this.apiClient.request<EmptyResponse>(
+      {
+        method: 'POST',
+        path: `/v3/connect/revoke`,
+        body: payload
+      },
+      {
+        responseSchema: EmptyResponseSchema,
```
agtang96, line 180: should this be HostedAuthResponseSchema or HostedAuthSchema? I don't see where those two schemas are used at all

b3ndoi: You are right, made the changes

mrashed-dev (approved):

PR #454, opened 2023-04-12
Delete returns 200 instead of 204 now

mrashed-dev:
# Description
The API was returning 204 for DELETE calls, now it returns 200 with a body.

# License

### DIFF
--- a/src/apiClient.ts
+++ b/src/apiClient.ts
@@ -4,6 +4,8 @@ import { NylasConfig, OverridableNylasConfig } from './config';
 import { NylasApiError, NylasAuthError } from './schema/error';
 import {
   AuthErrorResponseSchema,
+  DeleteResponse,
+  DeleteResponseSchema,
   ErrorResponseSchema,
 } from './schema/response';
 import { objKeysToCamelCase, objKeysToSnakeCase } from './utils';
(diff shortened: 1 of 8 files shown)


src/apiClient.ts
```diff
@@ -153,16 +155,21 @@ export default class APIClient {
 
   async requestWithResponse<T>(
     response: Response,
-    passthru: OptionsPassthru<ZodType<T>>
-  ): Promise<T> {
+    passthru?: OptionsPassthru<ZodType<T>>
+  ): Promise<T | DeleteResponse> {
```
agtang96, line 159: should just be T if the types are passed in correctly

src/apiClient.ts
```diff
@@ -217,10 +224,6 @@ export default class APIClient {
       );
     }
 
-    if (!passthru) {
```
agtang96, line 220: why?

src/apiClient.ts
```diff
@@ -172,15 +179,15 @@ export default class APIClient {
     );
   }
 
-  async request(options: RequestOptionsParams): Promise<undefined>;
```
agtang96, line 175: redundant

agtang96 (approved):

PR #469, opened 2023-06-16
Fixing an issue with Attirbute.Number transforming null values into 0

kraju3:
# Description
There was a bug that caused null values to return 0 for Attribute.Number. This will check if the value is null or not and return only if it's a valid Numbrer. This will avoid Number(null) returning 0.


mrashed-dev (approved):
LGTM!

Issue #496, opened 2023-10-18
Uncaught (in promise) Error: No access token in response


elcous:
**Describe the bug**
Error occurs when using this module in Deno environment

**To Reproduce**
Some steps involved to reproduce the bug and any code samples you can share.
```
import Nylas from "npm:nylas";

Nylas.config({
  clientId: Deno.env.get("CLIENT_ID"),
  clientSecret: Deno.env.get("CLIENT_SECRET"),
});

async function test() {
  const code = "123";

  Nylas.exchangeCodeForToken(code).then((token) => {
    console.log(token);
  });
}
```

**Expected behavior**
Error shouldn't occur

**SDK Version:**
6.10

**Additional context**
Most likely happens because of the bug here:
https://github.com/node-fetch/node-fetch/issues/1725

That was fixed. Probably can just upgrade node-fetch version to >2.6.12

mrashed-dev:
Thanks for opening this issue @elcous, just upgraded and the tests pass. Looks like we're good to upgrade!

PR #498, opened 2023-10-21
issue #496 Update node-fetch version to fix 'No access token in response' error

techwithsrikanth:
# License


mrashed-dev:
Sorry, but Node 3.x is not compatible with the current SDK because it does not support esm! Upgrading to 2.6.12 rectifies it.

PR #499, opened 2023-10-21
Undefined msg id code

techwithsrikanth:
updated the payload saving method to handle undefined messageid request

# License


mrashed-dev:
Sorry, but this code change is not required. The `?` beside the `replyToMessageId` denotes that it can be undefined so making it a union type with undefined is redundant.

PR #523, opened 2023-12-21
Fix issue with form-data not importing correctly for ESM projects

mrashed-dev:
# Description
The npm package `form-data` has an issue with importing compatibility between CJS and ESM. Using `import * as FormData` or `import FormData` works only on cjs while `import { default as FormData }` works only on esm. We work around this now by checking if `default` exists on the importing object, and using it if it does, otherwise just use `FormData`.

# License


relaxedtomato (approved):
thanks!

PR #524, opened 2024-01-03
Fix issue where Graph events were returning 400 on update

mrashed-dev:
# Description
This PR remedies the issue seen where Graph accounts were returning 400 on update due to visibility field being present in the PUT request. We now check to see if the visibility field was modified, and only then do we include it in the outgoing request.

# License


ajay-k (approved):
LGTM 👍

PR #551, opened 2024-03-22
chore: change the type property of Date, Datespan, Time and Timespan to object

### LINKED ISSUE #550: Type mismatch in Events - When DataType
**Describe the bug**
 So I came across this bug,  Inside the type of Event Object
 
 Context.
 
```
  const events = await nylas.events.list({
            identifier: account.grant_id,
            queryParams: queryParams,
          });
```  
          
` type When = Time | Timespan | Date | Datespan;`
 
 ```
export interface Time {
    /**
     * A UNIX timestamp representing the time of occurrence.
     */
    time: number;
    /**
     * If timezone is present, then the value for time will be read with timezone.
     * Timezone using IANA formatted string. (e.g. "America/New_York")
     * @see <a href="https://en.wikipedia.org/wiki/List_of_tz_database_time_zones">List of tz database time zones</a>
     */
    timezone: string;
    /**
     * The type of 'when' object.
     */
    type: WhenType.Time;
}

```
```
export interface Timespan {
    /**
     * The start time of the event.
     */
    startTime: number;
    /**
     * The end time of the event.
     */
    endTime: number;
    /**
     * The timezone of the start time.
     * Timezone using IANA formatted string. (e.g. "America/New_York")
     * @see <a href="https://en.wikipedia.org/wiki/List_of_tz_database_time_zones">List of tz database time zones</a>
     */
    startTimezone?: string;
    /**
     * The timezone of the end time.
     * Timezone using IANA formatted string. (e.g. "America/New_York")
     * @see <a href="https://en.wikipedia.org/wiki/List_of_tz_database_time_zones">List of tz database time zones</a>
     */
    endTimezone?: string;
    /**
     * The type of 'when' object.
     */
    type: WhenType.Timespan;
}
```

So problem aries as documentation for V3 Api mentions that it will return an

 "object" instead of "type" inside When dataType of Event.

```
ANY OF These
Timespan {
    start_time: number
    end_time: number
   start_timezone: string or null
   end_timezone: string or null
   object: string 🆁 timespan Default: timespan
}

 Date { 
      date: string
      object: string 🆁 Default: date
}
 
Datespan {
      start_date: string
      end_date: string
     object: string 🆁 Default: datespan
}
```
 
 In conclusion, there is a type mismatch between sdk & api doc. As api actually returns "object" instead of "type".
 

**To Reproduce**
Not sure? Just look at sdk type for Event
https://github.com/nylas/nylas-nodejs/blob/main/src/models/events.ts
https://developer.nylas.com/docs/api/v3/ecc/#get-/v3/grants/-grant_id-/events

 

**Expected behavior**
Need to update event.ts file to match api docs type.

**SDK Version:**
 "nylas": "^7.2.1",
 "I am using V3 api"

**Additional context**
I switched to sdk to avoid type mismatch between my codebase & nylas, if its fine then I can quickly update the types in repo and open PR.

hanut:
This PR updates the `type` property of the Date, Datespan, Time and Timespan types (the WhenTypes) to
`object` which is what is returned by the latest version (v3) of the Nylas api

# License

Closes #550

### DIFF
--- a/src/models/events.ts
+++ b/src/models/events.ts
@@ -511,7 +511,7 @@ export interface Datespan {
   /**
    * The type of 'when' object.
    */
-  type: WhenType.Datespan;
+  object: WhenType.Datespan;
 }
 
 /**
--- a/src/models/events.ts
+++ b/src/models/events.ts
@@ -443,7 +443,7 @@ export interface Time {
   /**
    * The type of 'when' object.
    */
-  type: WhenType.Time;
+  object: WhenType.Time;
 }
 
 /**
@@ -474,7 +474,7 @@ export interface Timespan {
   /**
    * The type of 'when' object.
    */
-  type: WhenType.Timespan;
+  object: WhenType.Timespan;
 }
 
 /**
@@ -490,7 +490,7 @@ export interface Date {
   /**
    * The type of 'when' object.
    */
-  type: WhenType.Date;
+  object: WhenType.Date;
 }
 
 /**


hanut:
> Looks good but can you make the same change to the Datespan interface? As soon as it's done I'll approve and merge 👍🏼 Thanks for jumping on this!

Thanks for pointing that out ! 
Always happy to help :+1:

PR #552, opened 2024-04-03
Add support for passing in headers to outgoing calls

mrashed-dev:
# Description
This PR allows you to pass in a map of additional, custom headers that can be sent to the Node SDK.

# Usage
This can be used one of two ways, for all outgoing calls or on a per-call-basis.

### All Calls
For all outgoing calls, you can set headers when initializing the Nylas SDK:
```js
const nylas = new Nylas({
  apiKey: "NYLAS_API_KEY",
  headers: {
    foo: "bar"
  },
});

messages = await nylas.messages.list({
  "identifier",
}); // Will be sent with 'foo: bar' header

calendars = await nylas.calendars.list({
  "identifier",
}); // Will be sent with 'foo: bar' header
```

### Per-call basis
For all calls to the API, we have an `override` object. This object is now extended to include a `headers` field.
```js
const nylas = new Nylas({
  apiKey: "NYLAS_API_KEY",
});

messages = await nylas.messages.list({
  "identifier",
  overrides: {
    headers: {
      foo: "bar"
    }
  }
}); // Will be sent with 'foo: bar' header

calendars = await nylas.calendars.list({
  "identifier",
}); // Will not be sent with 'foo: bar' header
```

# License

### DIFF
--- a/src/models/events.ts
+++ b/src/models/events.ts
@@ -354,10 +354,10 @@ type When = Time | Timespan | Date | Datespan;
  * Type representing the different objects representing time and duration when creating events.
  */
 type CreateWhen =
-  | Omit<Time, 'type'>
-  | Omit<Timespan, 'type'>
-  | Omit<Date, 'type'>
-  | Omit<Datespan, 'type'>;
+  | Omit<Time, 'object'>
+  | Omit<Timespan, 'object'>
+  | Omit<Date, 'object'>
+  | Omit<Datespan, 'object'>;
 
 /**
  * Enum representing the different types of when objects.
--- a/src/apiClient.ts
+++ b/src/apiClient.ts
@@ -65,11 +65,16 @@ export default class APIClient {
    * The timeout for requests to the Nylas API, in seconds
    */
   timeout: number;
+  /**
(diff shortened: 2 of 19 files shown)


src/models/events.ts
```diff
@@ -354,10 +354,10 @@ type When = Time | Timespan | Date | Datespan;
  * Type representing the different objects representing time and duration when creating events.
  */
 type CreateWhen =
-  | Omit<Time, 'type'>
-  | Omit<Timespan, 'type'>
-  | Omit<Date, 'type'>
-  | Omit<Datespan, 'type'>;
```
AaronDDM, line 360: Is this change expected?

AaronDDM (approved):
Could you add a test for some of the local overrides as well? Looks like we just have a test for the header being set at the top level.

src/models/events.ts
mrashed-dev: Yes this is related to #551, I think the tests never ran on that PR so this was uncaught.

mrashed-dev:
@AaronDDM good callout, tests added.

PR #554, opened 2024-04-09
Add support for include_grant_scopes=false on auth url generation

Taha-Sardar:
Add support for include_grant_scopes=false on auth url generation. Currently include grant scopes is true by default and is not set to false if the user sets the variable as false - This fix should enable the user to do so

# License


ashryanbeats:
Thanks for submitting @Taha-Sardar! I've requested the team to have a look.

PR #556, opened 2024-04-15
Add return type for send RSVP

mrashed-dev:
# Description
Send RSVP used to just return a `request_id` in the response, now it can return an error object in the event of a partial success where the send RSVP request is executed but the email was not sent.

# License


ozsivanov (approved):
LGTM

PR #559, opened 2024-04-26
Fixed issue where attachments < 3mb were not being encoded correctly

mrashed-dev:
# Description
This PR fixes an issue where we were not encoding attachments less than 3mb properly. Furthermore, for `CreateAttachmentRequest` you may now provide your own base64 encoded string if you prefer to not have the SDK encode it.

# License

### DIFF
--- a/src/resources/drafts.ts
+++ b/src/resources/drafts.ts
@@ -119,8 +119,8 @@ export class Drafts extends Resource {
 
     // Use form data only if the attachment size is greater than 3mb
     const attachmentSize =
-      requestBody.attachments?.reduce(function(_, attachment) {
-        return attachment.size || 0;
+      requestBody.attachments?.reduce((total, attachment) => {
+        return total + (attachment.size || 0);
       }, 0) || 0;
 
     if (attachmentSize >= Messages.MAXIMUM_JSON_ATTACHMENT_SIZE) {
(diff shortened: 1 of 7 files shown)


src/resources/messages.ts
```diff
@@ -210,7 +210,18 @@ export class Messages extends Resource {
     if (attachmentSize >= Messages.MAXIMUM_JSON_ATTACHMENT_SIZE) {
       requestOptions.form = Messages._buildFormRequest(requestBody);
     } else {
-      requestOptions.body = requestBody;
+      if (requestBody.attachments) {
```
VictorMonteiro7, line 213: As the documentation said:
_`To add attachments to a draft with a total HTTP payload size under 3MB, you can use either the application/json format or the multipart/form-data format.`_

The reduce above only gets the last attachment's size. 
I think the correct reduce would be:
```ts
 const maxLength = attachments.reduce((acc, attachment) => {
   acc += (attachment.size as number) || 0;
     return acc;
    }, 0) || 0;
```

mrashed-dev: 🤦🏼 Of course. Thanks for pointing this out, will rectify it 😄

src/resources/drafts.ts
```diff
@@ -131,6 +132,15 @@ export class Drafts extends Resource {
         form,
```
yifanplanet, line 132: Same comment goes for this:
```suggestion
  const attachmentSize =
      requestBody.attachments?.reduce((total, attachment) => {
        return total + (attachment.size || 0);
      }, 0) || 0;
```

src/resources/drafts.ts
```diff
@@ -167,6 +177,15 @@ export class Drafts extends Resource {
...
```
yifanplanet, line 177: And this:
```
  const attachmentSize =
    requestBody.attachments?.reduce((total, attachment) => {
      return total + (attachment.size || 0);
  }, 0) || 0;
```

yifanplanet (approved):
Just those two changes needed for the src/resources/drafts.ts

PR #564, opened 2024-05-15
Added support for filtering by event type when listing events

mrashed-dev:
# Description
This PR adds support for filtering events by event type. Note that this feature is only for Google events.

# License

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ### Unreleased
 * Added support for filtering by event type when listing events
+* Added support for filtering a list of folders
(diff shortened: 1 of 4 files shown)


kraju3 (commented):
@mrashed-dev Just curious when you pass in multiple eventTypes, do we join them together in the API request like this:

`event_type=default&event_type=outOfOffice`

https://developer.nylas.com/docs/api/v3/ecc/#get-/v3/grants/-grant_id-/events

mrashed-dev:
@kraju3 good call out, fixed it

kraju3 (approved):
LGTM

PR #565, opened 2024-05-16
CUST-2556 - Adding list folders query params

kraju3:
# Description 

Adding ListFolderQuery Params to the folders endpoint
https://developer.nylas.com/docs/api/v3/ecc/#get-/v3/grants/-grant_id-/folders

# License


mrashed-dev (changes requested):
@kraju3 looks good but I think you forgot to use `queryParams` in the list function itself.

PR #569, opened 2024-06-20
fix: application crash when timeout encountered

mmollick:
I encountered occasional application crashes and traced them back to timeouts coming from the Nylas API. After reviewing the Nylas-SDK, I found the culprit in the `setTimeout` call used to abort the request. It's bad practice to throw an error from `setTimeout` as its `try...catch` blocks will not intercept the error. Instead, these errors will crash the application.

`node-fetch` will reject when `controller.abort()` is called, so we don't need to throw from the timeout anyway.

Reference: 
https://jaketrent.com/post/catch-error-thrown-settimeout/
https://nodejs.org/dist/latest-v7.x/docs/api/errors.html#errors_node_js_style_callbacks

# License

### DIFF
--- a/src/apiClient.ts
+++ b/src/apiClient.ts
@@ -186,7 +186,7 @@ export default class APIClient {
 
       return response;
     } catch (error) {
-      if (error instanceof Error && error.name === "AbortError") {
+      if (error instanceof Error && error.name === 'AbortError') {
         throw new NylasSdkTimeoutError(req.url, this.timeout);
       }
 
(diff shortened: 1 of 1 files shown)


mrashed-dev (approved):
LGTM, thanks for your contribution!

src/apiClient.ts
```diff
@@ -187,6 +186,10 @@ export default class APIClient {
 
       return response;
     } catch (error) {
+      if (error instanceof Error && error.name === "AbortError") {
```
mrashed-dev, line 189: @mmollick is it possible for you to either run prettier (via `npm run lint:prettier`) or just use single quotes here so we can get the CI check to pass, it would be greatly appreciated!

```suggestion
      if (error instanceof Error && error.name === 'AbortError') {
```

mmollick: Done

PR #597, opened 2024-10-30
Fix the ability to rotate webhook secrets

### LINKED ISSUE #596: Unable to rotate webhook secrets
**Describe the bug**
The `webhooks.rotateSecret` method calls an incorrect endpoint, resulting in the call failing with an error.

```
[error]: Received an error but could not parse response from the server: Cannot POST /v3/webhooks/<webhookId>/rotate-secret
{
  "error": {},
  "requestId": "<requestId>",
  "user": {
    "id": "1",
    "email": "<email>"
  }
}
```

**To Reproduce**
```
const nylas = new Nylas({ apiKey });
const response = await nylasClient.webhooks.rotateSecret({ webhookId }); // -- fails here
```

**Expected behavior**
This method should call the correct endpoint, as defined in the documentation [here](https://developer.nylas.com/docs/api/v3/admin/#post-/v3/webhooks/rotate-secret/-id-):

> `POST /v3/webhooks/rotate-secret/{id}`

**SDK Version:**
7.6.1

**Additional context**
The endpoint described in the documentation works correctly when using an Axios instance to call the endpoint directly.

Zellotath:
# Description
The `webhooks.rotateSecret` method calls the incorrect URL, so all calls using this method fail with an error. This fixes that by switching the URL to the correct one. Closes #596. 

# License

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+### Unreleased
+* Fix for the webhook rotate secret calling the wrong endpoint
+
 ### 7.6.1 / 2024-10-30
 * Add support for filtering events by masterEventID
 * Add support for gzip compression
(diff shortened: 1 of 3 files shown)


SubashPradhan (approved):
Thanks so much for the fix.

PR #606, opened 2024-11-19
Fix query parameters for array values

wallacecs007:
We were creating a new snakeCaseKey=item for each item in the value array.

This was creating multiple instances of the snakeCaseKey query param if there is more than one item in an array.

If more than one value was passing more than one value in an array query parameter we were generating a URL that looks like this:
`/v3/grants/9a8ce69e-61a5-42c5-bdda-1ab3c3745093/threads?limit=5&any_email=chase.w%40nylas.com&any_email=nick.b%40nylas.com`

This change makes it so that the above URL instead produces the following request URL:
`/v3/grants/9a8ce69e-61a5-42c5-bdda-1ab3c3745093/threads?limit=5&any_email=chase.w%40nylas.com,nick.b%40nylas.com`

# License

### DIFF
--- a/src/apiClient.ts
+++ b/src/apiClient.ts
@@ -106,9 +106,7 @@ export default class APIClient {
           }
           url.searchParams.set('metadata_pair', metadataPair.join(','));
         } else if (Array.isArray(value)) {
-          for (const item of value) {
-            url.searchParams.append(snakeCaseKey, item as string);
-          }
+          url.searchParams.append(snakeCaseKey, value.join(','));
         } else if (typeof value === 'object') {
           for (const item in value) {
             url.searchParams.append(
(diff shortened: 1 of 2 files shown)


src/apiClient.ts
```diff
@@ -106,9 +106,7 @@ export default class APIClient {
           }
           url.searchParams.set('metadata_pair', metadataPair.join(','));
         } else if (Array.isArray(value)) {
-          for (const item of value) {
-            url.searchParams.append(snakeCaseKey, item as string);
-          }
+          url.searchParams.append(snakeCaseKey, value.join(','));
```
tjovicic, line 109: @wallacecs007 can you write a test? Also, will this apply only to the `anyEmail` query param or all?

wallacecs007: @tjovicic I fixed the test that was failing due to this change and this will apply to all query parameters. I don't see any case in our API spec docs where we would want multiple of the same query parameter passed in to Nylas outside of Metadata.

Metadata has it's own case and will be unaffected.

PR #623, opened 2025-02-27
Add support for listImportEvents method to import events from calendars

AaronDDM:
# Background
This adds support for the new beta endpoint [GET /v3/grants/{grant_id}/events/import](https://developer.nylas.com/docs/api/v3/ecc/#get-/v3/grants/-grant_id-/events/import), which allows importing events from a calendar within a specified time frame. This is useful for applications that need to synchronize events with their own calendaring solution or enrich events with custom data.

# License

### DIFF
--- a/src/models/events.ts
+++ b/src/models/events.ts
@@ -336,7 +336,7 @@ export type SendRsvpQueryParams = FindEventQueryParams;
 /**
  * Interface representing the query parameters for importing events.
  */
-export interface ImportEventQueryParams {
+export interface ListImportEventQueryParams {
   /**
    * (Required) Calendar ID to import events from. "primary" is a supported value indicating the user's primary calendar.
    * Note: "primary" is not supported for iCloud.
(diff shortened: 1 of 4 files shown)


src/models/events.ts
```diff
@@ -333,6 +333,27 @@ export type DestroyEventQueryParams = CreateEventQueryParams;
...
+   * Defaults to the time of the request.
+   */
+  start?: number;
+  /**
+   * Filter for events that end at or before the specified time, in Unix timestamp format.
+   * Defaults to one month from the time of the request.
+   */
+  end?: number;
+}
+
```
kraju3, line 356: I think we are missing the `select` parameter here

src/resources/events.ts
```diff
@@ -74,6 +75,15 @@ export interface DestroyEventParams {
   queryParams: DestroyEventQueryParams;
 }
 
+/**
+ * @property identifier The identifier of the grant to act upon
+ * @property queryParams The query parameters to include in the request
+ */
+export interface ListImportEventParams {
+  identifier: string;
+  queryParams: ListImportEventQueryParams;
```
kraju3, line 84: `page_token` and limit included here?

src/models/events.ts
AaronDDM: It should be inherited from `ListQueryParams`

kraju3 (approved):

Issue #633, opened 2025-04-17
Override timeout treated as milliseconds


yulianovdey:
**Describe the bug**
In https://github.com/nylas/nylas-nodejs/pull/617 the SDK was updated to respect a timeout passed as an override, but as a result the SDK now handles timeout values passed to the `APIClient` vs overrides differently:
- `timeout` passed in the constructor to `Nylas` is treated as seconds
- `timeout` passed in `overrides` for a request is treated as milliseconds

They both refer to `NylasConfig` so the param is documented as seconds in both cases.

Additionally, if a timeout error is caught the SDK always logs the client timeout, not the override if one was provided:

```typescript
      if (error instanceof Error && error.name === 'AbortError') {
        throw new NylasSdkTimeoutError(req.url, this.timeout);
      }
```

**To Reproduce**

```typescript
import Nylas from "nylas";

const nylas = new Nylas({
  apiKey: process.env.NYLAS_V3_API_KEY!,
  timeout: 5, // will be interpreted as seconds
});

(async () => {
  // This will succeed - request takes less than 5 seconds
  await nylas.threads.find({
    identifier: 'grant_id',
    threadId: 'thread_id',
  });

  // This will fail with NylasSdkTimeoutError, and incorrectly log "timeout: 5000"
  await nylas.threads.find({
    identifier: 'grant_id',
    threadId: 'thread_id',
    overrides: {
      timeout: 7, // will be interpreted as milliseconds
    },
  });
})();
```

**Expected behavior**
Ideally both timeouts are treated the same way. Otherwise, the documentation is updated to reflect that overrides are treated differently.

**SDK Version:**
7.8.0

AaronDDM:
Fixed in #634

PR #634, opened 2025-04-17
[Node SDK] consistent handling of timeout values for constructor and overrides

ajay-k:
# Description
This pull request fixes an issue where request timeout overrides were inconsistently handled compared to the timeout set in the constructor. Previously, the timeout in the constructor was treated as seconds and converted to milliseconds, while timeout in request overrides was incorrectly treated directly as milliseconds.

The changes in this PR ensure that:
Consistent behavior with backward compatibility:
- Values < 1000 are treated as seconds and converted to milliseconds internally (new behavior)
- Values ≥ 1000 are still treated as milliseconds for backward compatibility
- A deprecation warning is shown for values ≥ 1000, pointing users toward using seconds in the next major release

**Correct error reporting:**
Error messages correctly display the actual timeout value that was used for the request (in seconds)
Previously it would always show the client timeout regardless of which timeout was used
These changes maintain API compatibility while fixing the inconsistent behavior, providing a clear migration path for users. After the next major version, all timeout values will be treated consistently as seconds.

- Starts to address [#633](https://github.com/nylas/nylas-nodejs/issues/633)
- [Jira Ticket](https://nylas.atlassian.net/browse/CUST-4296)
- We'll need to address the change and remove the warning message in the next release, this fix signals to the customer to prepare for this change and adds a warning message

# License


ajay-k:
After discussing this with @AaronDDM we both agree implementing this right away would be a breaking change, we're going to handle this by:

- Figuring out if the customer is using milliseconds and add a warning to let customers know it's being deprecated in next major release and to use seconds instead
- Had Cursor AI write some tests for me to make sure it worked as expected:

> 🔍 TEST 1: Small timeout value (< 1000, treated as seconds)
>   Making request with override timeout of 5 seconds...
> (node:84791) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
> (Use `node --trace-deprecation ...` to show where the warning was created)
>   ✅ Request succeeded with small timeout value
>   First thread ID: 19639fa7f8a200b8
> 
> 🔍 TEST 2: Large timeout value (≥ 1000, treated as milliseconds)
>   Making request with override timeout of 3000 milliseconds...
>   ✅ Request succeeded with large timeout value
>   First thread ID: 19639fa7f8a200b8
>   ✅ Deprecation warning shown as expected:
>   "DEPRECATED: Providing timeout in milliseconds via request overrides is deprecated and will be removed in the next major release. Please use seconds instead."
> 
> 🔍 TEST 3: Force timeout error with very small value (0.001 seconds)
>   Making request with tiny timeout to force an error...
>   ✅ Timeout error occurred as expected
>   Error details:
>     Type: NylasSdkTimeoutError
>     Message: Nylas SDK timed out before receiving a response from the server.
>     Timeout: 0.001 seconds
>     URL: https://api.us.nylas.com/v3/grants/8eab7c11-9143-406d-928f-57b16ca45d1e/threads?limit=1

src/apiClient.ts
```diff
@@ -156,7 +156,25 @@ export default class APIClient {
...
+    
+    // Handle timeout backward compatibility
+    let timeoutDuration: number;
+    if (options.overrides?.timeout) {
+      // Determine if the override timeout is likely in milliseconds (≥ 1000)
+      if (options.overrides.timeout >= 1000) {
+        // eslint-disable-next-line no-console
+        console.warn(
+          'DEPRECATED: Providing timeout in milliseconds via request overrides is deprecated and will be removed in the next major release. Please use seconds instead.'
+        );
```
AaronDDM, line 168: @ajay-k let's remove this console warn message and instead add a `@deprecated [message here]` to the `options.overrides.timeout` property defined in the `OverridableNylasConfig` interface/type. So we'll just use typescript to notify of this impending deprecation.

Also update the CHANGELOG.md to call out this deprecation.

Finally, you have some github actions failing that need to be resolved.

Thank you!

ajay-k: Hey @AaronDDM made some changes you suggested, can you review and lmk if it looks good to you as well?

AaronDDM (approved):

PR #650, opened 2025-06-27
Adding sending via raw mime

kraju3:
# License

### DIFF
--- a/src/models/messages.ts
+++ b/src/models/messages.ts
@@ -356,3 +356,28 @@ export interface CleanMessagesResponse extends Message {
    */
   conversation: string;
 }
+
+/**
+ * Interface representing a request to send a message using raw MIME format.
+ */
+export interface SendMimeMessageRequest {
+  /**
+   * The raw MIME content of the message to send.
+   */
+  mime: string;
+  /**
+   * Optional metadata to attach to the message.
+   * Defaults to empty string if not provided.
+   */
+  metadata?: string;
+}
+
+/**
+ * Interface representing the query parameters for sending a MIME message.
+ */
+export interface SendMimeMessageQueryParams {
+  /**
+   * The type of message being sent. Must be 'mime' for MIME messages.
+   */
+  type: 'mime';
+}
--- a/src/resources/messages.ts
+++ b/src/resources/messages.ts
@@ -291,15 +291,15 @@ export class Messages extends Resource {
     const path = makePathParams('/v3/grants/{identifier}/messages/send', {
       identifier,
     });
-    
+
     // Create FormData for MIME message
     const FD = require('form-data');
     const FormDataConstructor = FD.default || FD;
     const form: FormData = new FormDataConstructor();
-    
+
     // Add MIME content
     form.append('mime', requestBody.mime);
-    
+
     // Add metadata (defaults to empty string if not provided)
     if (requestBody.metadata) {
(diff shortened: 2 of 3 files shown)


samuelpx (approved):

AaronDDM (changes requested):
Thanks for adding raw‑MIME send support — the feature itself is welcome. I'd like a few things addressed before this is merged.

**1. The branch is significantly behind `main` and won't compile after rebase.**
- `mergeable_state` is `dirty` (conflicts), and on the branch `package.json` still lists `form-data ^4.0.0` and `node-fetch ^2.6.12` as dependencies. On `main` those have been removed; the SDK now uses `formdata-node` + `form-data-encoder` (see `package.json` and `src/resources/messages.ts:1` on `main`: `import { Blob, FormData } from 'formdata-node';`). Please rebase onto `main`, then switch the implementation over to the project's existing form‑data setup (see point 2).

**2. Use the project's existing `FormData` import, not a dynamic `require('form-data')`.**
In `src/resources/messages.ts` the new `sendMime` method does:
```ts
const FD = require('form-data');
const FormDataConstructor = FD.default || FD;
const form: FormData = new FormDataConstructor();
```
This is inconsistent with the rest of the file (which uses `formdata-node`'s `FormData` via static ESM import) and the `form: FormData` annotation is the DOM type, not the runtime instance. After rebase, just reuse the top‑level `FormData` import and drop the `require` / `.default || FD` fallback. The dynamic require will also break the bundled output if `form-data` isn't a project dependency.

**3. Caller shouldn't have to pass `queryParams: { type: 'mime' }`.**
The method is literally named `sendMime`; making the caller supply `{ type: 'mime' }` is redundant and easy to get wrong. Set it inside the method instead and remove `SendMimeMessageQueryParams` from the public API. (Compare the existing `send()` shape — callers shouldn't need to remember magic strings the SDK can supply.)

**4. `metadata` doc/behavior mismatch.**
In `src/models/messages.ts`, the JSDoc on `SendMimeMessageRequest.metadata` says *"Defaults to empty string if not provided."* but in `src/resources/messages.ts` the implementation does:
```ts
if (requestBody.metadata) {
  form.append('metadata', requestBody.metadata);
}
```
That's not an empty‑string default — it just skips the field entirely when falsy. Either change the doc to say "omitted if not provided" or actually append `''` when undefined. Also note `if (requestBody.metadata)` will swallow legitimate empty strings if the API ever expected one.

**5. Duplicate `beforeEach` in tests.**
`tests/resources/messages.spec.ts` adds a second `beforeEach` that duplicates the existing one immediately above (`jest.clearAllMocks(); apiClient.request.mockResolvedValue({});`). Please drop the duplicate.

**6. Missing CHANGELOG entry.**
There's no `Unreleased` entry in `CHANGELOG.md` for this new public method. Add one (something like `Added messages.sendMime() for sending raw MIME messages`).

**7. Worth discussing: top‑level `sendMime` vs. a `send()` option.**
The v3 API exposes this as the same `/messages/send` endpoint with `type=mime`. Other SDKs may want to follow the convention of overloading `send()` with a `type` flag rather than adding a parallel method. Worth syncing with whoever's owning consistency across the Python/Ruby/Java SDKs before locking the surface area down.

Requesting changes mainly for the rebase + form‑data import + redundant `type: 'mime'` items.

PR #672, opened 2025-09-25
Fix Nylas Node.js SDK issue 669

AaronDDM:
This pull request contains changes generated by Cursor background composer.


AaronDDM:
> Not just the values, but the keys too. Can we do an audit of all other apis to ensure this is correct?

All of our models use camel case which is translated to snake case to match the expected API schemas.

We are working on having our models automatically generated from our OpenAPI spec files, which will resolve such inconsistencies.

radenkovic (approved):

PR #736, opened 2026-06-11
TW-5371 Add admin API SDK resources

AaronDDM:
## Summary

Updates the Node SDK admin/application API surface to match source-code truth for Applications, Redirect URIs, Manage Domains, Workspaces, Policies, Rules, and grant rule evaluations.

## Changes

- Add Manage Domains resource for `/v3/admin/domains` CRUD, `info`, and `verify`.
- Add Workspaces resource for `/v3/workspaces` CRUD, `autoGroup`, and `manualAssign`.
- Add `applications.update()` for `PATCH /v3/applications`.
- Correct redirect URI update to `PATCH` and delete return typing.
- Normalize the nested `/v3/rules` list envelope from `data.items` / `data.nextCursor`.
- Update models, tests, and CHANGELOG.

## Jira

- https://nylas.atlassian.net/browse/TW-5371

## Validation

- `npm test -- --coverage=false --runTestsByPath tests/resources/applications.spec.ts tests/resources/redirectUris.spec.ts tests/resources/policies.spec.ts tests/resources/rules.spec.ts tests/resources/domains.spec.ts tests/resources/workspaces.spec.ts`
- `npm run build --if-present`
- Independent model verifier (`codex exec`, default `gpt-5.5`) PASS for endpoint/list-shape accuracy. Fable 5 CLI attempt was blocked by local Claude session limit, so Codex was used as the requested model-like verifier fallback.


qasim-nylas (approved):
LGTM 👍

qasim-nylas (approved):
LGTM 👍

src/resources/domains.ts
```diff
@@ -0,0 +1,200 @@
...
+ * UUID or a domain address (FQDN/email format).
+ */
+interface FindDomainParams {
+  domainId: string;
+}
+
+/**
+ * @property requestBody The values to create the domain with.
+ */
+interface CreateDomainParams {
```
sourcesoft, line 42: The /v3/admin/domains uses Nylas Service Account for authentication and requires request singing. The request is very unique/different:
```
curl -X POST "https://api.us.nylas.com/v3/admin/domains" \
  -H "Content-Type: application/json" \
  -H "X-Nylas-Signature: <BASE64_SIGNATURE>" \
  -H "X-Nylas-Kid: <SERVICE_ACCOUNT_ID>" \
  -H "X-Nylas-Nonce: <NONCE>" \
  -H "X-Nylas-Timestamp: 1742932766" \
  -d '{
    "name": "My transactional domain",
    "domain_address": "mail.example.com"
  }'
```
Do we first support this type of authentication? Read more here: https://developer.nylas.com/docs/v3/auth/nylas-service-account/

sourcesoft: If we already support this, then we're good. So just leaving this as a comment for confirmation

AaronDDM: Confirmed. The Node SDK does not generate Nylas Service Account signatures automatically, but these Manage Domains methods now require and pass through the four signed request headers (`X-Nylas-Kid`, `X-Nylas-Timestamp`, `X-Nylas-Nonce`, `X-Nylas-Signature`) via request overrides. So callers can use their existing signing implementation and the SDK will send the unique auth headers unchanged.

PR #746, opened 2026-06-19
Add missing `message.deleted` Webhook trigger

Kile:
Hi there. I noticed that your `WebhookTrigger` enum doesn't include `message.deleted` which, while left out in a bunch of lists of possible triggers in the docs, still exist as a possible trigger:
https://developer.nylas.com/docs/reference/notifications/messages/message-deleted/

This isn't a huge thing at all but I am quite obsessed with types so it would be cool if this was in the enum and I didn't have to do
```ts
type WebhookTriggersWithDeleted = WebhookTriggers | 'message.deleted';
```

# License

### DIFF
--- a/src/models/webhooks.ts
+++ b/src/models/webhooks.ts
@@ -133,6 +133,7 @@ export enum WebhookTriggers {
   // Message triggers
   MessageCreated = 'message.created',
   MessageUpdated = 'message.updated',
+  MessageDeleted = 'message.deleted',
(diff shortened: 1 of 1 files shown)


radenkovic:
@Kile thank you for the contribution, I've added an elaborate PR with all the missing webhooks so this one is also covered.

Kile:
Thank you, makes more sense in a bigger PR 🙏

PR #754, opened 2026-07-02
TDT-2791 Clarify event visibility default is Google-only

samLRodrigues:
## What

Updates the JSDoc for the `Visibility` type in `src/models/events.ts` to note that `default` only works for Google events. Microsoft and EWS accounts return a 400 error when `default` is sent, but the type declaration didn't warn about this.

Jira: TDT-2791 (sub-task of TDT-2765)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


AaronDDM (changes requested):
Good afternoon! Thanks for tightening this up — the note is accurate and the provider-specific 400 behaviour is worth calling out. One thing before this ships: there's no entry under the `Unreleased` section in `CHANGELOG.md`. Please add a short line under `Changed` (e.g. "Clarify that event `default` visibility is Google-only") so the doc fix is tracked. Docs themselves look good.

AaronDDM (approved):

PR #759, opened 2026-08-26
NYL-22: Add Contact metadata and webhook parity

quzhi1:
## Summary

- add `Contact.metadata` request/response typing and `metadataPair` contact filtering
- document metadata replace, preserve, and clear behavior plus provider webhook support boundaries
- verify `contact.updated` and `contact.deleted` wire values and guard against unsupported `contact.created`

## Test plan

- `npm test -- --coverage=false` (488 tests passed)
- `npm run build`
- Prettier on changed files


src/models/contacts.ts
```diff
@@ -118,6 +123,12 @@ export interface ListContactQueryParams extends ListQueryParams {
    * When set to true, returns the contacts also within the specified Contact Group subgroups, if the group parameter is set.
    */
   recurse?: boolean;
+  /**
+   * Filters contacts by one indexed metadata entry in `key:value` form.
+   * Use one of `key1` through `key5`. This filter cannot be combined with
+   * provider-side contact filters; pagination parameters are supported.
+   */
+  metadataPair?: Record<string, string>;
```
AaronDDM, line 131: Heads up on a cross-SDK difference: `apiClient.ts` (~line 110) joins all `metadataPair` entries into a single comma-separated `metadata_pair=k1:v1,k2:v2`, while Java/Python/Ruby emit a repeated `metadata_pair` param per entry. The doc says "one indexed metadata entry", but nothing stops a caller passing two — and if they do, Node sends a different wire shape than the other SDKs. Not introduced here, just worth knowing since this is the first contacts consumer.

tests/resources/contacts.spec.ts
```diff
@@ -44,6 +44,23 @@ describe('Contacts', () => {
       });
     });
 
+    it('should pass an indexed metadata filter to the API', async () => {
```
AaronDDM, line 47: This asserts the params reach `apiClient.request`, not what actually goes on the wire. The serialization is the interesting part (see comment on contacts.ts) — the Java PR added a URL-level assertion in `NylasClientTest`. Worth one here too.

AaronDDM (approved):
Good morning! Nice clean change — types, docs, changelog and tests are all here, and the `contact.created` guard test is a good touch.

Two notes below, neither blocking. Approving.

src/models/contacts.ts
quzhi1: Acknowledged. The generic multi-entry join behavior predates NYL-22, while the Contact contract accepts exactly one indexed pair. I kept that shared serializer stable and added a Contact-path URL assertion for the supported single-pair wire shape in 881a3c4.

tests/resources/contacts.spec.ts
quzhi1: Added a URL-level assertion in `tests/apiClient.spec.ts` that verifies a Contact list filter becomes `metadata_pair=key1:sync_eligible` on the outgoing URL. Included in 881a3c4.

AaronDDM (approved):
Good morning! Re-reviewed after 881a3c4.

Both items from yesterday are addressed:
- `tests/apiClient.spec.ts` now asserts the real outgoing URL (`metadata_pair=key1%3Async_eligible`), which was the gap.
- The multi-entry join in `apiClient.ts` is pre-existing shared-serializer behavior and out of scope here — agreed on leaving it.

CHANGELOG has an Unreleased entry, doc comments cover the replace/preserve/clear semantics, and `UpdateContactRequest` picks up `metadata` via the alias to `CreateContactRequest`. No new concerns.

Approving.

